PR #29, opened 2026-02-06
feat(tui): integrate self-update into Settings

waylon256yhw:
## Summary

- Add non-blocking update check/download/apply to TUI Settings, built on top of the existing `commands/update.rs` CLI update infrastructure
- Introduce a narrow `pub(crate)` API (`check_for_update`, `download_and_apply`, `UpdateCheckInfo`) so TUI code never touches internal types like `ReleaseAsset`
- Add optional progress callback to `download_release_asset` (64KB throttle); CLI path passes `None`, TUI path receives live progress

## Implementation

**update.rs**: 2 new `pub(crate)` async functions + 1 struct. `download_release_asset` gains an `Option<&dyn Fn(u64, Option<u64>)>` parameter — existing CLI call site passes `None`, no behavior change.

**mod.rs**: `UpdateSystem` worker thread (pattern matches existing `SpeedtestSystem`/`WebDavSystem`). Worker holds `last_tag` internally to avoid leaking version info into `Action`/`Overlay` enums. Channel send errors are handled to prevent overlay getting stuck.

**app.rs**: 3 `Overlay` variants (`UpdateAvailable`, `UpdateDownloading`, `UpdateResult`), 3 `Action` variants, `SettingsItem::CheckForUpdates`. Keyboard handling: arrow keys toggle buttons, Enter confirms, Esc cancels/hides. Successful update sets `should_quit = true`.

**ui.rs**: Settings row shows current version. 3 overlay renderers — version comparison with button pair, `Gauge` progress bar, success(green)/error(red) result.

**i18n.rs**: 15 bilingual (zh/en) text functions.

## Test compatibility

- `download_release_asset` signature change is internal (`fn`, not `pub`), and the only existing call site is updated to pass `None` — **all 18 upstream update unit tests pass unchanged**
- New `pub(crate)` functions are async + require network, consistent with the existing pattern of no unit tests for network-dependent update paths
- `cargo test` full suite passes

## Test plan

- [x] `cargo build` compiles cleanly
- [x] `cargo test` — all existing tests pass (including 18 update tests)
- [x] Manual: Settings → Check for Updates → shows "already latest" toast
- [x] Manual: With older version → UpdateAvailable overlay → button navigation works
- [x] Manual: Confirm update → download progress gauge → UpdateResult → auto-exit
- [x] Manual: Esc during download → overlay hidden → background toast on completion
- [x] Manual: Language switch → all update texts render correctly in zh/en

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,23 @@ All notable changes to CC Switch CLI will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [4.7.1] - 2026-02-07
+
+### Added
+
+- **Providers (TUI)**: Add RightCode provider template.
+
+### Changed
+
+- **TUI**: Tighten layout spacing and reduce list left padding.
+- **Codex (TUI)**: Split auth/config previews for clearer navigation.
+- **Docs**: Update home screenshots (EN/ZH).
+
+### Fixed
+
+- **Codex**: Align provider config and auth handling with upstream.
+- **TUI**: Add left padding between nav border and icons.
+
 ## [4.7.0] - 2026-02-07
 
 ### Added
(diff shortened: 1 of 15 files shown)


SaladDay:
@waylon256yhw thanks🩷

PR #75, opened 2026-03-23
fix(claude): preserve user-defined settings on provider switch

zj1123581321:
## Summary

- When switching providers, `write_claude_live()` previously overwrote the entire `settings.json`, causing user-defined fields to be lost
- Now preserves known user-configurable top-level keys (`statusLine`, `hooks`, `permissions`, `apiKeyHelper`) from the existing file when writing new provider config
- This fixes the issue where HUD plugins (e.g. `claude-hud`) stop working after a provider switch because `statusLine` config is dropped

## Details

The root cause is in `write_claude_live()` (`src-tauri/src/services/provider/claude.rs`): it constructs the new config from the provider's `settings_config` + common config snippet, then writes it directly to `settings.json` without reading the existing file. Any fields not stored in the provider's database record are lost.

The fix reads the existing `settings.json` before writing and carries over user-defined fields that are not present in the new provider config. A whitelist approach (`PRESERVE_KEYS`) is used instead of preserving all unknown keys, to avoid conflicts with common config management (e.g. clearing a common config snippet should still remove its fields).

## Test plan

- [x] All 946+ existing tests pass (0 failures)
- [x] Manual test: switch claude-official → zhipu → claude-official, `statusLine` preserved throughout
- [x] Manual test: `env` correctly switches between providers
- [x] The previously failing test `clear_updates_live_config_even_without_apply_flag` passes (common config fields are correctly removed when cleared)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src-tauri/src/services/provider/claude.rs
+++ b/src-tauri/src/services/provider/claude.rs
@@ -235,7 +235,7 @@ impl ProviderService {
         let mut provider_content = provider.settings_config.clone();
         let _ = Self::normalize_claude_models_in_value(&mut provider_content);
 
-        let content_to_write = if let Some(snippet) = common_config_snippet {
+        let new_content = if let Some(snippet) = common_config_snippet {
             let snippet = snippet.trim();
             if snippet.is_empty() {
                 provider_content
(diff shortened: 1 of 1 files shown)


SaladDay:
Superseded by the live-config rework on main. Instead of preserving a fixed key set (statusLine/hooks/permissions/apiKeyHelper) during the write, those now sit in the common-config snippet and get reapplied on each switch, so there's no per-write merge to keep in sync. Closing — the explicit key list isn't needed under the new model. Appreciate the report.

PR #117, opened 2026-04-19
Fix model settings not switching during temp launch

yuanzidev:
The `start claude` command only extracted the `env` sub-object from `settings_config` when building the temporary settings file. This meant model overrides (e.g. `ANTHROPIC_DEFAULT_SONNET_MODEL`) and other top-level fields were missing from the temp file, causing stale model names from the global `~/.claude/settings.json` to bleed through when Claude CLI merged `--settings` with its default config.

Now `prepare_launch_with` writes the full `settings_config` (cloned and model-normalized via `normalize_claude_models_in_value`), matching the behavior of the `provider switch` path.


SaladDay:
THANKS！！🙂

PR #131, opened 2026-04-25
fix: preserve unmanaged Claude settings on switch

KiBlazer:
## Summary

  - Preserve unmanaged fields in `~/.claude/settings.json` when switching Claude providers.
  - Keep provider-owned Claude env keys managed by cc-switch-cli, removing stale provider values during live writes.
  - Prevent live-only Claude settings from being backfilled into provider snapshots.

  ## Test Plan

  - `cargo fmt --check`
  - `cargo test --lib`
  - Real switch test: `chatgpt-ep -> chatgpt-k1587-163 -> chatgpt-ep`
    - verified custom live-only test fields were preserved
    - verified provider-owned `ANTHROPIC_*` values switched correctly
    - verified exported provider snapshot did not include live-only test fields

### DIFF
--- a/src-tauri/src/services/provider/claude.rs
+++ b/src-tauri/src/services/provider/claude.rs
@@ -1,6 +1,19 @@
+use super::common::merge_json_values;
 use super::*;
 
 impl ProviderService {
+    const CLAUDE_PROVIDER_ENV_KEYS: &'static [&'static str] = &[
+        "ANTHROPIC_AUTH_TOKEN",
+        "ANTHROPIC_API_KEY",
+        "ANTHROPIC_BASE_URL",
+        "ANTHROPIC_MODEL",
+        "ANTHROPIC_REASONING_MODEL",
+        "ANTHROPIC_SMALL_FAST_MODEL",
+        "ANTHROPIC_DEFAULT_HAIKU_MODEL",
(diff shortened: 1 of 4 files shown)


SaladDay:
感谢你提交这个修复，方向是很有价值的 👍

我这边看了一下，有几个地方需要再调整一下：

1. 现在 PR 和最新 main 有冲突，尤其是 #123 fix(provider): unify common config live handling 已经重构了 common config 的写入/剥离语义，需要先 rebase 到最新 main 后重新处理冲突。
2. 需要尽量复用当前 main 里的 common_config::* 统一逻辑，不要回退到旧的 Claude-only merge/strip 路径。特别是 common config 的 precedence 现在已经语义调整过。
3. “保留 unmanaged live settings” 这个思路是对的，但建议把它做成 Claude live write 外层的窄修复：只保护 statusLine、permissions 等 Claude 用户配置，同时继续让 cc-switch 管理 provider-owned 的 ANTHROPIC_* 字段。

SaladDay:
main reworks the live-config write path, so this ends up handled a different way. The merge/conflict step on switch is gone — switching writes the selected provider directly, and shared/non-provider settings are carried in the common-config snippet instead of being merged back each time. Closing since the approach here is superseded, but thanks for the writeup.

PR #144, opened 2026-04-29
Fix proxy failover status output

### LINKED ISSUE #143: CLI proxy output incorrectly says automatic failover is disabled
## Summary

The CLI proxy status/output currently reports that automatic failover is disabled even when per-app proxy config has `auto_failover_enabled = true`.

Observed output from `cc-switch proxy show` / `cc-switch proxy serve` includes messages such as:

```text
Mode: manual takeover only (automatic failover disabled)
Manual takeover only. Automatic failover is disabled in this phase.
```

## Why this looks incorrect

The proxy request path already reads `auto_failover_enabled` and uses the failover queue when it is enabled:

- `ProviderRouter::select_providers` reads `get_proxy_config_for_app(app_type).auto_failover_enabled`.
- When enabled, it selects providers from `get_failover_queue(app_type)`.
- `HandlerContext` and `ForwardOptions` use the app proxy config to enable timeouts and avoid bypassing the circuit breaker.

So the hard-coded CLI message can make users think failover is unavailable even when the runtime is configured to use it.

## Expected behavior

CLI output should reflect the actual per-app automatic failover setting, or use neutral wording that points users to the app proxy settings and failover queue.

## Reproduction shape

1. Configure an app with `auto_failover_enabled = true` and a failover queue.
2. Run `cc-switch proxy show` or `cc-switch proxy serve`.
3. The CLI still says automatic failover is disabled.

## Proposed fix

Update CLI proxy output to show per-app automatic failover status instead of hard-coding it as disabled.

apple-ouyang:
## Summary

Fixes #143.

This updates the CLI proxy output so it no longer hard-codes automatic failover as disabled. The proxy runtime already reads per-app `auto_failover_enabled` and routes through the configured failover queue when enabled, so the old message was misleading.

Changes:

- Show a neutral local-proxy mode description in `cc-switch proxy show`.
- Print per-app automatic failover status for Claude, Codex, and Gemini.
- Make `cc-switch proxy serve` print the configured automatic failover status instead of saying it is disabled.
- Add a regression test that enables Codex auto failover and verifies the CLI status output reflects it.

## Tests

- `cargo fmt --all`
- `cargo test cli::commands::proxy::tests::proxy_overview_lines --lib`
- `cargo test proxy::provider_router::tests --lib`

Note: `cargo test --lib` on my local macOS environment still has two unrelated proxy runtime/session failures that reproduce when run individually. One fails because a local cc-switch process is already listening on `127.0.0.1:15721`; the other is `managed_session_ready_info_accepts_persisted_session_without_status_probe`.

### DIFF
--- a/src-tauri/src/cli/commands/proxy.rs
+++ b/src-tauri/src/cli/commands/proxy.rs
@@ -156,13 +156,9 @@ fn serve_proxy(
                     ))
                 );
             }
-            println!(
-                "{}",
-                info(crate::t!(
-                    "Manual takeover only. Automatic failover is disabled in this phase.",
-                    "仅支持手动接管；本阶段不包含自动故障转移。"
-                ))
-            );
+            for line in build_auto_failover_status_lines(&state) {
+                println!("{}", info(&line));
+            }
             println!(
                 "{}",
                 info(crate::t!(
@@ -284,8 +280,8 @@ fn build_proxy_overview_lines(
             listen_port
         ),
         crate::t!(
-            "Mode: manual takeover only (automatic failover disabled)",
-            "模式：仅支持手动接管（自动故障转移已禁用）"
+            "Mode: local proxy (manual takeover and automatic failover follow app settings)",
+            "模式：本地代理（手动接管和自动故障转移遵循各应用配置）"
         )
         .to_string(),
         format!(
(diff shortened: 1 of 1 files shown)


unive3sal:
Hi @apple-ouyang 

Thanks for your contribution! Would you please rebase your commit for review?

apple-ouyang:
Rebased, should be OK now. Please take another look.

SaladDay:
LGTM, cc @unive3sal

SaladDay (approved):

PR #152, opened 2026-05-03
feat: respect CLAUDE_CONFIG_DIR env var for Claude Code

LuJiansen:
## Summary

Add support for `CLAUDE_CONFIG_DIR` environment variable in cc-switch. This allows users to direct Claude Code config files to a custom directory instead of the default `~/.claude`, matching the behavior of the Claude Code CLI's own `CLAUDE_CONFIG_DIR` env var.

Only Claude Code is covered in this PR. The other 4 apps (Codex, Gemini, OpenCode, OpenClaw) are intentionally excluded — their env var names would be invented rather than official, and may conflict with future upstream XDG support.

## Changes

- **Claude**: `CLAUDE_CONFIG_DIR` (already officially supported by Claude Code CLI)
- Other apps: use cc-switch's existing settings override mechanism instead

## Priority chain

```
CLAUDE_CONFIG_DIR env var > cc-switch settings override > $HOME/.claude
```

## Why

On shared Linux accounts (multiple users on one UID), all config files land in the same `~/.claude`. Setting `CLAUDE_CONFIG_DIR` allows each user to point to their own private directory, keeping API tokens and provider state isolated — without needing filesystem permissions or wrapping the binary.

## Test plan

- [x] Unit tests for Claude config dir with env var set, blank, and unset
- [x] Existing test suite passes unchanged
- [x] Manual verification: provider list, provider switch with `CLAUDE_CONFIG_DIR` set
- [x] Backward compatible — no env var → behavior unchanged

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -24,13 +24,21 @@ const CODEX_RESERVED_MODEL_PROVIDER_IDS: &[&str] = &[
     "ollama-chat",
 ];
 
-/// 获取 Codex 配置目录路径
+/// Returns the Codex config directory path.
+///
+/// Priority: `CODEX_CONFIG_DIR` env var > cc-switch settings override > `$HOME/.codex`
 pub fn get_codex_config_dir() -> PathBuf {
+    if let Some(dir) = std::env::var_os("CODEX_CONFIG_DIR") {
+        let dir = PathBuf::from(dir);
+        if !dir.as_os_str().is_empty() && !dir.to_string_lossy().trim().is_empty() {
+            return dir;
+        }
+    }
     if let Some(custom) = crate::settings::get_codex_override_dir() {
         return custom;
     }
 
-    home_dir().expect("无法获取用户主目录").join(".codex")
+    home_dir().expect("Could not determine home directory").join(".codex")
 }
 
 /// 获取 Codex auth.json 路径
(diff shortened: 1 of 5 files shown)


unive3sal:
Hi @LuJiansen 

Thanks for your contribution. But I strongly suggest you **only** keep the XDG adapter for claude code.
Because we have no idea if other agents will officially support XDG configuration directory in the future. Once they supports XDG directory but use a different environment name, the cc-switch-cli will mismatch with it

LuJiansen:
Hi @unive3sal ,
Thanks for your suggestion, I have remove XDG adapter for other apps.

SaladDay:
@LuJiansen  I think this is worth merging after a small cleanup.

  - Keep the diff scoped to Claude only.
  - Drop unrelated comment/message changes in Codex/Gemini/OpenCode/OpenClaw files.
  - Add tests for priority: CLAUDE_CONFIG_DIR should override cc-switch settings, and blank CLAUDE_CONFIG_DIR should fall back to the settings override.

  After that, I’m in favor of merging it.

LuJiansen:
@SaladDay , I have made corresponding revisions, thanks!

SaladDay:
LGTM, Thanks!

SaladDay (approved):

PR #155, opened 2026-05-03
feat(tui): add failover controls

### LINKED ISSUE #19: 故障转移
是否可以添加和原版一样的故障转移功能

unive3sal:
A huge commit for failover TUI and integration tests. It is mainly consist of the following parts:

* user can manually enable/disable auto-failover
* provider SHOULD ONLY selected from failover queue when failover is enable
* user can specify a provider from provider list when failover is disable (use "space" key to specify the active provider)
* user cannot specify one typical provider but set the priority in failover queue when failover is enable 
(all of the provider in queue SHOULD be marked as "✓" status)
* if (failover is disable && user does not specify a provider), then proxy cannot be activated
* if (failover is enable && failover queue is empty), then proxy cannot be activate 
* failover queue should support hot updates during proxy enabled
(but cannot remove all providers, it should always keep at least 1 provider in queue)

# interactive failover TUI

add a temp window in "Providers"

Screenshot 2026-05-03 at 23 36 18

## when failover is disable

*  user can specify ONLY 1 provider from provider list

## when failover is enable

* user can update priority for each provider

Screenshot 2026-05-03 at 23 39 13

# E2E integration test

* It will create a sandbox to run E2E tests
* Creating some positive/negative mock providers during test

## edge cases

* failover queue is empty; expectation: cannot enable proxy
* failover queue has ONLY 1 provider; expectation: ONLY use this provider in queue. even if it is down, active provider cannot be transfer
* failover queue has more than 1 provider; expectation: auto-transfer to different provider according to priority when current active provider is unavailable.
* all of the providers in queue are unavailable; expectation: cannot work normally

Fixes #19
Fixes #101 
Fixes #98

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -44,3 +44,30 @@ jobs:
       - name: Check formatting
         working-directory: src-tauri
         run: cargo fmt --check
+
+  failover-e2e:
+    name: failover E2E test
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Rust
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: 1.91.1
+
+      - name: Setup Rust cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: src-tauri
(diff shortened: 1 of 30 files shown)


SaladDay (approved):
merge一下main

PR #159, opened 2026-05-06
Fix Codex official login and temp auth persistence

TMYTiMidlY:
## Goal

This PR fixes the `openai-official` Codex provider flow in two parts.

### 1. Switching to `openai-official` should not break Codex TUI startup

Before this PR, switching to the OpenAI official provider could write an empty live `~/.codex/auth.json`. That is different from having no auth file at all: Codex sees the file, tries to load account credentials from it during TUI bootstrap, and then fails with the account/read error instead of entering its normal login flow.

This PR treats an empty auth snapshot for an official Codex provider as "no saved login". When applying `openai-official`, cc-switch removes the live `~/.codex/auth.json` instead of writing `{}`. With no live auth file present, Codex follows its normal official login path, so the TUI no longer trips over an empty auth file during startup.

### 2. Temporary `cc-switch start codex openai-official` login state should be written back

`cc-switch start codex openai-official` launches Codex with an isolated temporary `CODEX_HOME` seeded from the selected provider. This avoids mutating the user's global Codex home while the temporary start session is running.

After Codex exits, cc-switch captures useful state from that temporary home back into the selected provider. If the user logs in during the temporary start session, the temp `auth.json` is saved back into the `openai-official` provider so the next start can reuse it.

If the provider was already logged in and the user logs out during the temporary start session, Codex removes the temp `auth.json`. The capture step records that logout state in the provider, and later applying or starting `openai-official` keeps `auth.json` absent rather than writing an empty `{}` file. This preserves logout without reintroducing the TUI startup error.

## Summary

- Remove live `~/.codex/auth.json` when switching to an official Codex provider with an empty auth snapshot, so Codex can prompt for login normally instead of failing during TUI startup.
- Launch `cc-switch start codex openai-official` with a temporary `CODEX_HOME` seeded from the selected provider snapshot.
- After the Codex handoff exits, capture temp `auth.json` and `config.toml` back into the selected provider snapshot.
- Persist a missing temp `auth.json` as logout state, and make later live/temp application delete or omit `auth.json` rather than writing an empty `{}` file.
- Strip runtime MCP tables from captured temp `config.toml` before saving it back into the provider.
- Bypass startup recovery for the hidden internal capture command so unrelated startup recovery failures cannot lose captured login state.
- Use a hidden `internal` command name instead of `__internal` to avoid Bash completion generation panics.
- Keep user-facing error messages localized, with Chinese text in the zh branch and English text in the en branch.

## Tests

- `cargo fmt`
- `cargo test codex_temp_launch`
- `cargo test internal_commands_bypass_future_schema_database_gate`
- `cargo test commands_that_skip_startup_state_do_not_touch_future_schema_database`
- `cargo test completions`
- `cargo run --quiet -- completions bash`
- `cargo build --release`
- `/home/timidly/.local/bin/cc-switch --version`
- `/home/timidly/.local/bin/cc-switch completions bash`

### DIFF
--- a/src-tauri/src/cli/codex_temp_launch.rs
+++ b/src-tauri/src/cli/codex_temp_launch.rs
@@ -12,6 +12,8 @@ use serde_json::Value;
 #[derive(Debug, Clone)]
 pub(crate) struct PreparedCodexLaunch {
     pub(crate) executable: PathBuf,
+    pub(crate) cc_switch_executable: PathBuf,
+    pub(crate) provider_id: String,
     pub(crate) codex_home: PathBuf,
 }
 
@@ -37,9 +39,18 @@ where
     Resolve: FnOnce() -> Result<PathBuf, AppError>,
 {
     let executable = resolve_codex_binary()?;
+    let cc_switch_executable = std::env::current_exe().map_err(|err| {
+        AppError::localized(
+            "codex.temp_launch_current_exe_failed",
+            format!("解析当前 cc-switch 可执行文件路径失败: {err}"),
+            format!("Failed to resolve current cc-switch executable path: {err}"),
+        )
+    })?;
     let codex_home = write_temp_codex_home(temp_dir, provider)?;
     Ok(PreparedCodexLaunch {
         executable,
+        cc_switch_executable,
+        provider_id: provider.id.clone(),
         codex_home,
     })
 }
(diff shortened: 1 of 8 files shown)


SaladDay:
LGTM, THANKS!

TMYTiMidlY:
Thanks for the quick review and merge! Really appreciate the fast turnaround.

PR #167, opened 2026-05-07
Fix broken internal documentation links

aqilaziz:
## Summary
- point Chinese release note language links to the existing English files in docs
- replace stale CLAUDE.md references with existing README documentation

## Verification
- checked all local Markdown links
- git diff --check


SaladDay:
LGTM, THANKS

Issue #182, opened 2026-05-16
[feature request] add vim-style navigation support for form_handler


feiyehua:
Currently in "add provider" page it doesn't support vim-style navigation, which is anti-intuitive.

<img width="910" height="397" alt="Image" src="https://github.com/user-attachments/assets/a689850c-08df-416b-9df3-35fc2fb8b8da" />
I've created a branch to settle this problem at https://github.com/feiyehua/cc-switch-cli/tree/feiyehua/form-vim-navigation, hope this can be merged

Lan-Fa:
bro knows me😭😭😭

PR #183, opened 2026-05-16
feat(tui) add vim-style navigation in form handler

feiyehua:
As discussed in #182, add vim-style navigation support for forms (such as when focusing "fields" during adding a provider)
image


SaladDay:
LGTM, THANKS

PR #188, opened 2026-05-18
feat(tui) normalize vim-style navigation across forms and overlays

feiyehua:
Follow-up to #183.

This extends vim-style navigation handling across TUI forms and overlays by normalizing `hjkl` at the main key-entry point when no text input is active, while preserving literal `hjkl` input in editing contexts.

Changes:
- add shared editing-state checks for forms and overlays
- route `hjkl` as navigation only when text input/editor/filter/form editing is inactive
- remove duplicated manual `j/k` handling from individual form handlers
- add regression coverage for form navigation and overlay text-input behavior


SaladDay:
LGTM, THANKS

PR #195, opened 2026-05-20
feat: improve model filling and overlay restore in model picker

Paulkm2006:
## Summary

- Add `fill all` option when setting model for Claude Code
- Fix overlay restore for model fetch picker
- Fix Enter key in model fetch picker to select first model when input is empty

## Changes

- Added tui_key_fill_all(), tui_claude_model_label_for_index(), tui_claude_model_fill_all_title(), tui_claude_model_fill_all_message(), and tui_claude_model_fill_all_empty_source() i18n functions in `src-tauri/src/cli/i18n.rs`
- Added `ClaudeModelFillAll { source_idx: usize }` variant to ConfirmAction enum in `src-tauri/src/cli/tui/app/types.rs`
- Added `a` key handler in `src-tauri/src/cli/tui/app/overlay_handlers/pickers.rs` — in `ClaudeModelPicker` select mode, pressing a opens a confirmation dialog showing which field will be the source; if the source field is empty, a toast warning is shown instead
- Added confirmation handling in `src-tauri/src/cli/tui/app/overlay_handlers/dialogs.rs` — Enter/y fills all 5 model fields from the source and restores `ClaudeModelPicker`; Esc/n cancels and restores `ClaudeModelPicker` without changing values
- Updated the key bar in `src-tauri/src/cli/tui/ui/overlay/pickers.rs` to display the a key hint
- In `handle_model_fetch_picker_key` Enter handler (`src-tauri/src/cli/tui/app/overlay_handlers/pickers.rs`), when the field is `ClaudeModelConfig`, the overlay now restores to `Overlay::ClaudeModelPicker` instead of `Overlay::None`
- Similarly, the Esc key also returns to `ClaudeModelPicker` when the source was the Claude model picker
- When pressing Enter in `ModelFetchPicker` with an empty input, it now selects the first filtered model instead of closing the overlay

## Tests

- Added 11 focused tests in `src-tauri/src/cli/tui/app/tests.rs` covering: a opens confirm without changing values, confirm fills all fields (Enter and y), cancel leaves values unchanged (Esc and n), empty source is a no-op, fill from non-first field, Claude `ModelFetchPicker` Enter/Esc restores to ClaudeModelPicker, Enter with empty input picks first model, and Hermes `ModelFetchPicker` restores via `pending_overlay`
- Passed the following unit tests
  - cargo fmt --manifest-path src-tauri/Cargo.toml -- --check
  - cargo test --manifest-path src-tauri/Cargo.toml claude_model -- --nocapture
  - cargo test --manifest-path src-tauri/Cargo.toml model_fetch_picker -- --nocapture
- Tested functions locally


SaladDay:
I think it is possible to make a focused operation row, like Fill All open/apply, where the user enters to execute? This key feels a bit out of place.

Paulkm2006:
> I think it is possible to make a focused operation row, like Fill All open/apply, where the user enters to execute? This key feels a bit out of place.

Thanks for the feedback. I considered adding "Fill All" as a dedicated row, but I think the current a key approach works better for this use case:

The user needs to choose which model to fill from. With the `a` key, the workflow is intuitive: navigate to the row you want, press a, and that field's value gets copied to the other 4 fields. 

However, if "Fill All" were a separate row, the source model would need to be implicit (e.g., "first non-empty") or we need to pop up another model picking overlay, which could be confusing and less efficient.

That being said, if you believe that the dedicated row approach is better than my current proposal, I'm happy to adjust the PR accordingly.

SaladDay:
Thanks for explaining the `a` key flow. I think keeping `a` is acceptable here because the currently focused row gives a clear source model, but this should still be guarded as a batch overwrite action.

Blocking items before merge:

- Please add a confirmation dialog before `a` actually fills all Claude model fields. The confirmation should make the source explicit, e.g. "Fill all Claude model fields from <field label>: <value>? Existing values will be overwritten." Pressing `a` should not mutate the form until the user confirms.
- If the selected source field is empty, please do not fill all fields with an empty value. A toast/no-op is safer.
- The current `ModelFetchPicker` restore change regresses the Hermes model picker flow. For non-Claude fields, `Enter`/`Esc` now set `self.overlay = Overlay::None` directly and bypass `close_overlay()`, which drops `pending_overlay`. Hermes model fetch depends on `pending_overlay = Some(Overlay::HermesModelsPicker { editing: false })`, so after selecting a fetched Hermes model we should return to the Hermes models picker. The existing test `model_fetch_picker_hermes_models_sets_current_model_id_only` catches this regression.
- Please add focused tests for the new behavior: `a` opens confirmation without immediately changing values, confirm fills all fields, cancel leaves values unchanged, empty source is a no-op, Claude model fetch restores to `ClaudeModelPicker`, and Hermes model fetch still restores to `HermesModelsPicker`.

I ran the targeted checks locally:

- `cargo fmt --manifest-path src-tauri/Cargo.toml -- --check` passed.
- `cargo test --manifest-path src-tauri/Cargo.toml claude_model -- --nocapture` passed.
- `cargo test --manifest-path src-tauri/Cargo.toml model_fetch_picker -- --nocapture` failed on `model_fetch_picker_hermes_models_sets_current_model_id_only`, matching the overlay restore regression above.

Paulkm2006:
> Thanks for explaining the `a` key flow. I think keeping `a` is acceptable here because the currently focused row gives a clear source model, but this should still be guarded as a batch overwrite action.
> 
> Blocking items before merge:
> 
> * Please add a confirmation dialog before `a` actually fills all Claude model fields. The confirmation should make the source explicit, e.g. "Fill all Claude model fields from : ? Existing values will be overwritten." Pressing `a` should not mutate the form until the user confirms.
> * If the selected source field is empty, please do not fill all fields with an empty value. A toast/no-op is safer.
> * The current `ModelFetchPicker` restore change regresses the Hermes model picker flow. For non-Claude fields, `Enter`/`Esc` now set `self.overlay = Overlay::None` directly and bypass `close_overlay()`, which drops `pending_overlay`. Hermes model fetch depends on `pending_overlay = Some(Overlay::HermesModelsPicker { editing: false })`, so after selecting a fetched Hermes model we should return to the Hermes models picker. The existing test `model_fetch_picker_hermes_models_sets_current_model_id_only` catches this regression.
> * Please add focused tests for the new behavior: `a` opens confirmation without immediately changing values, confirm fills all fields, cancel leaves values unchanged, empty source is a no-op, Claude model fetch restores to `ClaudeModelPicker`, and Hermes model fetch still restores to `HermesModelsPicker`.
> 
> I ran the targeted checks locally:
> 
> * `cargo fmt --manifest-path src-tauri/Cargo.toml -- --check` passed.
> * `cargo test --manifest-path src-tauri/Cargo.toml claude_model -- --nocapture` passed.
> * `cargo test --manifest-path src-tauri/Cargo.toml model_fetch_picker -- --nocapture` failed on `model_fetch_picker_hermes_models_sets_current_model_id_only`, matching the overlay restore regression above.

Hi, I've pushed new commits and updated the PR comment. This should fix the problems you mentioned before.

SaladDay:
LGTM, THANKS

PR #198, opened 2026-05-21
Fix Claude model fetch picker returning to model picker

feiyehua:
**Description**

This PR fixes the Claude provider model selection flow so that returning from the model fetch picker preserves the Claude model picker context.

Previously, opening model fetch from the Claude model picker and then confirming or cancelling would close the overlay entirely. Now the fetch picker remembers the originating Claude model picker row and returns to `ClaudeModelPicker` when appropriate.

**Changes**

- Added `model_picker_selected_row` to `Overlay::ModelFetchPicker` to restore previous ClaudeModelPicker.
- Captured the selected Claude model row when launching model fetch from `ClaudeModelPicker`.
- Restored `ClaudeModelPicker` on `Esc`, empty `Enter`, and successful `Enter` when the fetch picker originated there.
- Kept non-Claude/direct model fetch behavior unchanged.
- Added regression tests for Enter, Esc, empty input, and correct Claude model slot updates.

**Testing**

```bash
cargo test model_fetch_picker --manifest-path src-tauri/Cargo.toml
```

### DIFF
--- a/src-tauri/src/cli/tui/app/overlay_handlers/pickers.rs
+++ b/src-tauri/src/cli/tui/app/overlay_handlers/pickers.rs
@@ -351,6 +351,7 @@ impl App {
         let Overlay::ModelFetchPicker {
             field,
             claude_idx,
+            model_picker_selected_row,
             input,
             query,
             models,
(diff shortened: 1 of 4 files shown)


feiyehua:
close as duplicate to #195

PR #202, opened 2026-05-23
feat: duplicate provider configuration

feiyehua:
## Summary

- Add `c` keybinding in the TUI provider list to quickly duplicate an existing provider, pre-filling a new add form with the original's settings and a generated unique ID
- Implement the CLI `duplicate_provider` command (previously a stub)

## Changes

- **TUI provider list**: pressing `c` opens an add form pre-populated from the selected provider, with a new name (`{name} copy`), a generated ID, and identity/timestamp fields cleared
- **CLI**: `duplicate_provider` now reads the original provider, generates a unique ID, strips identity fields, and saves via `ProviderService::add`
- **i18n**: added `tui_key_copy()` text (中文: "复制", English: "copy")
- **Tests**: added `provider_copy_form_uses_new_record_identity_without_queue_state` to verify the copy form produces correct output

### DIFF
--- a/src-tauri/src/cli/commands/provider.rs
+++ b/src-tauri/src/cli/commands/provider.rs
@@ -448,39 +448,13 @@ fn edit_provider(app_type: AppType, id: &str) -> Result<(), AppError> {
 
 fn duplicate_provider(app_type: AppType, id: &str) -> Result<(), AppError> {
     let state = AppState::try_new()?;
-    let (original, existing_ids) = {
-        let config = state.config.read().map_err(AppError::from)?;
-        let manager = config
-            .get_manager(&app_type)
-            .ok_or_else(|| AppError::Message(texts::app_config_not_found(app_type.as_str())))?;
-        let original = manager
-            .providers
-            .get(id)
-            .ok_or_else(|| {
-                let msg = texts::entity_not_found(texts::entity_provider(), id);
-                AppError::localized("provider.not_found", msg.clone(), msg)
-            })?
-            .clone();
-        let existing_ids = manager.providers.keys().cloned().collect::<Vec<_>>();
-        (original, existing_ids)
-    };
-
-    let duplicate_name = format!("{} copy", original.name.trim());
-    let duplicate_id = generate_provider_id(&duplicate_name, &existing_ids);
-    let mut duplicate = original;
-    duplicate.id = duplicate_id.clone();
-    duplicate.name = duplicate_name;
-    duplicate.created_at = None;
-    duplicate.sort_index = None;
-    duplicate.in_failover_queue = false;
-
-    ProviderService::add(&state, app_type, duplicate)?;
+    let duplicate = ProviderService::duplicate(&state, app_type, id, None)?;
 
     println!(
         "{}",
         success(&format!(
             "✓ Duplicated provider '{}' as '{}'",
-            id, duplicate_id
+            id, duplicate.id
         ))
     );
     Ok(())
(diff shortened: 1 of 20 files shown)


SaladDay:
Thanks for the PR. The feature direction is useful, but I don’t think this is ready to merge yet. A few behavior gaps should be treated as blocking:

- **Backend behavior should stay aligned with upstream duplicate semantics.**  
  Upstream already has provider duplication, and the backend/TUI behavior here should mirror it as closely as practical. In particular, for OpenCode / OpenClaw / Hermes, upstream creates the duplicate without immediately writing it into the live app config (`addToLive = false`) and checks live provider IDs to avoid collisions. In this PR, the CLI path calls `ProviderService::add` directly, which treats additive apps as live/current and can write the duplicate into the target app config. It also only checks IDs from the local manager, not live-only provider IDs.

- **Copied providers should preserve upstream sort behavior.**  
  The PR clears `sortIndex`, so the duplicate can fall to the end of the list. Upstream inserts the copied provider directly below the original provider and shifts later providers as needed. Please keep that behavior for TUI/CLI parity.

- **The `c` shortcut is acceptable, but it should guard against accidental copies.**  
  I’m okay with using `c` for copy in the provider list, but since this creates a new provider, please show a confirmation popup before opening/creating the duplicate. That keeps the shortcut efficient while avoiding accidental copies.

- **Tests should cover the actual duplicate workflow.**  
  The added form-level test is useful, but it doesn’t cover live-write behavior, live-only ID collision handling, sort position, or the new confirmation flow. Please add focused tests for those cases.

Small cleanup: `src-tauri/src/cli/commands/provider.rs` now has an unused `error` import warning.

feiyehua:
> Thanks for the PR. The feature direction is useful, but I don’t think this is ready to merge yet. A few behavior gaps should be treated as blocking:
> 
> * **Backend behavior should stay aligned with upstream duplicate semantics.**
>   Upstream already has provider duplication, and the backend/TUI behavior here should mirror it as closely as practical. In particular, for OpenCode / OpenClaw / Hermes, upstream creates the duplicate without immediately writing it into the live app config (`addToLive = false`) and checks live provider IDs to avoid collisions. In this PR, the CLI path calls `ProviderService::add` directly, which treats additive apps as live/current and can write the duplicate into the target app config. It also only checks IDs from the local manager, not live-only provider IDs.
> * **Copied providers should preserve upstream sort behavior.**
>   The PR clears `sortIndex`, so the duplicate can fall to the end of the list. Upstream inserts the copied provider directly below the original provider and shifts later providers as needed. Please keep that behavior for TUI/CLI parity.
> * **The `c` shortcut is acceptable, but it should guard against accidental copies.**
>   I’m okay with using `c` for copy in the provider list, but since this creates a new provider, please show a confirmation popup before opening/creating the duplicate. That keeps the shortcut efficient while avoiding accidental copies.
> * **Tests should cover the actual duplicate workflow.**
>   The added form-level test is useful, but it doesn’t cover live-write behavior, live-only ID collision handling, sort position, or the new confirmation flow. Please add focused tests for those cases.
> 
> Small cleanup: `src-tauri/src/cli/commands/provider.rs` now has an unused `error` import warning.

Thanks for the comments. 
for the first problem, cc-switch-cli 's provider add interface now doesn't handle additive-apps correctly (namely, only reads managed configs), maybe these problems should be handled together?
For the second problem, cc-switch-cli now leaves `sortIndex` NULL, and the duplication(copy) behavior matches this. If we need to match the upstream, maybe we need to further rewrite the provider add function.
Tests and guard added.
PS: hermes_config::tests::set_current_provider_writes_model_section fails, seems caused by incorrectly managed lock

SaladDay:
LGTM, THANKS

PR #206, opened 2026-05-26
feat: add session management TUI

SaladDay:
## Summary
- add upstream-aligned session scanning/loading/deletion support for local Claude/Codex/Gemini/Hermes/OpenClaw/OpenCode sessions
- add a Sessions TUI page with app-scoped lists, lazy message loading, resume/delete shortcuts, and anchored relative timestamps
- keep session work on a background worker with stale-result guards and focused regression tests

## Verification
- cargo fmt --manifest-path src-tauri/Cargo.toml -- --check
- cargo test --manifest-path src-tauri/Cargo.toml sessions --lib
- cargo test --manifest-path src-tauri/Cargo.toml session_req_drain --lib
- cargo test --manifest-path src-tauri/Cargo.toml session_terminal_target_matches_upstream_setting_names --lib
- cargo test --manifest-path src-tauri/Cargo.toml tui_sessions_list_time_column_uses_relative_time_before_date --lib
- cargo test --manifest-path src-tauri/Cargo.toml relative_time_matches_upstream_thresholds --lib
- git diff --check

Fixes #142

PR #214, opened 2026-05-28
feat(tui): add session message filters

shuiqingliu:
## Summary
- add a Sessions message role filter (`f`) that cycles All/User/AI/Tool/Developer/System
- extend Sessions search so loaded messages can be filtered by role, localized role label, content, or timestamp
- keep selection/navigation stable when the current role/search filter changes

## Reference
Addresses feedback from #206 requesting session filters such as “用户” and search support to find a prior question and its AI answer.

## Tests
- `cargo +1.91.1 test sessions_ --lib -- --nocapture`
- `cargo +1.91.1 check --all-targets`

### DIFF
--- a/src-tauri/src/cli/i18n.rs
+++ b/src-tauri/src/cli/i18n.rs
@@ -7426,6 +7426,18 @@ pub mod texts {
         }
     }
 
+    pub fn tui_sessions_messages_title_with_filter(query: Option<&str>) -> String {
+        let mut title = tui_sessions_messages_title().to_string();
+        if let Some(query) = query.filter(|value| !value.trim().is_empty()) {
+            if is_chinese() {
+                title.push_str(&format!(" · 搜索: {}", query.trim()));
+            } else {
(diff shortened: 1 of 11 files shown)


SaladDay:
Maybe you can take a look at how to integrate with "/"?

shuiqingliu:
Sure, let me try it. I'll take a look at the code and update the PR soon.

shuiqingliu:
Updated. `/` integration is now implemented.

- In Sessions, `/` follows the active pane:
  - List pane: filters sessions.
  - Message pane: filters messages in the selected session.
- This supports the combined flow: search sessions by `AI`, switch to Messages, then search `User` there without losing the session filter.
- Exact message role queries like `User` / `用户` match the message role instead of assistant text that happens to mention “user”.

Validation:
- `cargo +1.91.1 test sessions_ --lib -- --nocapture`
- `cargo +1.91.1 check --all-targets`

SaladDay:
LGTM,thanks

PR #226, opened 2026-06-01
feat(deeplink): add `deeplink` command importing provider/mcp/prompt/skill

LeonardoTan19:
## Summary

- Port MCP/prompt/skill import handlers from upstream (farion1231/cc-switch), adding library-level deeplink import for resource types beyond provider.
- Wire the ported handlers into the deeplink parser and expose a top-level `cc-switch deeplink <url>` command that dispatches by `resource` type (provider/mcp/prompt/skill), with the deep link's own `app`/`apps` parameters driving the target.
- Align imported-prompt timestamps to seconds to match the prompt store (upstream used milliseconds).
- Add integration tests covering MCP server persistence with merged app flags, prompt import with enabled/disabled states, skill repo registration, and malformed-repo rejection.

## Test plan

- [x] `cargo test deeplink` passes all integration tests
- [x] `cargo fmt --check` and `cargo clippy` clean


LeonardoTan19:
@SaladDay  `--app` 是个全局的参数，跟deeplink不是很搭，应该对其采取什么样的处理方式呢？

SaladDay:
@LeonardoTan19 I think the cleanest behavior is to make the deeplink URL the single source of truth for target apps. So `--app` should not override the URL, and we should not silently ignore it either.

If users run something like:

```bash
cc-switch --app codex deeplink "ccswitch://v1/import?resource=provider&app=claude..."
```

we should fail fast with a clear message, for example:

```text
`--app` cannot be used with `deeplink`; target app(s) must be encoded in the URL via `app` or `apps`.
```

That keeps the protocol reproducible for scripts/geeks, and avoids surprising behavior for beginners.

LeonardoTan19:
@SaladDay Thanks for your suggestion! I agree on it. I'll get this implemented as soon as possible.

SaladDay:
Wowwww!! Thanks you for your contribution!!

LeonardoTan19:
I have fully replicated the cc-switch behavior for handling deeplinks. This command enables CLI-based deeplink imports. Since it reuses utils.rs, prompt.rs, etc., pulled directly from cc-switch, it shares the exact same Base64 decoding logic—meaning the issue with `%20`(blankspace) parsing errors is also reproduced here. Hope this bug can be solved in the future.

While I believe we could allow deeplinks to directly trigger the TUI interface in the future, that is out of scope for this PR.

This is now ready for review!

SaladDay:
@LeonardoTan19 Thanks for the update. I checked this against upstream `farion1231/cc-switch`, and the library-level deeplink implementation looks aligned with upstream overall. The CLI command wrapper is the only cc-switch-cli-specific part, since upstream handles deeplinks through Tauri commands instead.

There is one small blocker before we can merge: `cargo test --test deeplink_import --no-run` currently fails because `deeplink_import.rs` still calls `deeplink::execute(cmd)` with the old signature. Since `execute` now takes the global app flag as the second argument, that test should call:

```rust
cc_switch_lib::cli::commands::deeplink::execute(cmd, None)
```

After that, please run `cargo test --test deeplink_import`. Once it passes, I think this PR should be good to merge.

LeonardoTan19:
> @LeonardoTan19 Thanks for the update. I checked this against upstream `farion1231/cc-switch`, and the library-level deeplink implementation looks aligned with upstream overall. The CLI command wrapper is the only cc-switch-cli-specific part, since upstream handles deeplinks through Tauri commands instead.
> 
> There is one small blocker before we can merge: `cargo test --test deeplink_import --no-run` currently fails because `deeplink_import.rs` still calls `deeplink::execute(cmd)` with the old signature. Since `execute` now takes the global app flag as the second argument, that test should call:
> 
> ```rust
> cc_switch_lib::cli::commands::deeplink::execute(cmd, None)
> ```
> 
> After that, please run `cargo test --test deeplink_import`. Once it passes, I think this PR should be good to merge.

Done, I've fixed it.

SaladDay:
LGTM, THANKS!!!

Issue #222, opened 2026-05-31
Reuse temp launch flow for session restore instead of external terminal resume


feiyehua:
The Sessions restore flow currently launches an external terminal and runs a plain resume command.

This path has several implementation problems around terminal launching, quoting, working directory handling, and behavior consistency across terminal apps. It also differs from the existing Claude temporary launch flow, which already provides a better handoff model for starting Claude from the TUI.

Claude temp launch already handles the launch lifecycle more appropriately for this use case. Session restore should reuse that behavior instead of going through the generic external-terminal resume path.

**Expected Behavior**

When restoring a session from Sessions, reuse the existing temp launch flow, change the working directory and pass the session UUID as native Claude arguments:

```sh
--resume <UUID>
```

The restore path should not launch Claude through the generic external terminal resume implementation.

It should also avoid introducing a separate temp settings mechanism.

PR #233, opened 2026-06-04
fix(tui): resume sessions in current terminal (#222)

feiyehua:
Replace the external terminal launch approach for session resume with direct `exec` in the current terminal.

Previously, resuming a session would attempt to open an external terminal application (Terminal.app, iTerm2, Ghostty, etc.) via platform-specific launchers. 
Now, session resume uses `exec` (on Unix) to replace the current process with the resume command in the user's shell, giving a seamless in-terminal experience. On non-Unix platforms, the resume command is shown as a fallback for manual execution.

### Changes

- **Remove `session_manager::terminal` module** — deletes all platform-specific terminal launchers (macOS Terminal, iTerm2, Ghostty, Kitty, WezTerm, Kaku, Alacritty, Warp, custom)
- uses `CommandExt::exec()` to replace the current process with the user's shell running the resume command
- **Deprecate `preferred_terminal` setting** — field is retained for serde compatibility but marked `#[allow(dead_code)]`

I didn't add confirmation before launch agent apps to align with the behavior of existing "launch temp"


SaladDay:
Thanks for working on this. I do not think this PR fixes #222 yet.

The issue asks the Sessions restore path to reuse the existing Claude temp launch flow and pass `--resume <UUID>` as native Claude arguments. That temp launch path builds the effective provider snapshot, writes the temporary Claude settings file, and hands off as `claude --settings <temp> ...`.

This PR removes the external terminal launcher, but `Action::SessionResume` still just executes the saved `resume_command` through the user shell (`$SHELL -c`). That means a Claude session restore can still run against whatever live Claude config is currently on disk, instead of the provider selected in cc-switch. In practice, this keeps the main bug class from #222: restoring a session may not use the current provider configuration.

I would not merge this as-is. The better fix is to make the session resume action provider-aware, carry `provider_id`, `session_id`, and `project_dir`, and for Claude reuse the existing temp launch implementation with native args like `--resume <session_id>` while setting the session cwd. The current-provider temp settings machinery should stay single-sourced there, rather than adding or preserving a generic shell-command resume path for Claude.

feiyehua:
> Thanks for working on this. I do not think this PR fixes #222 yet.
> 
> The issue asks the Sessions restore path to reuse the existing Claude temp launch flow and pass `--resume <UUID>` as native Claude arguments. That temp launch path builds the effective provider snapshot, writes the temporary Claude settings file, and hands off as `claude --settings <temp> ...`.
> 
> This PR removes the external terminal launcher, but `Action::SessionResume` still just executes the saved `resume_command` through the user shell (`$SHELL -c`). That means a Claude session restore can still run against whatever live Claude config is currently on disk, instead of the provider selected in cc-switch. In practice, this keeps the main bug class from #222: restoring a session may not use the current provider configuration.
> 
> I would not merge this as-is. The better fix is to make the session resume action provider-aware, carry `provider_id`, `session_id`, and `project_dir`, and for Claude reuse the existing temp launch implementation with native args like `--resume <session_id>` while setting the session cwd. The current-provider temp settings machinery should stay single-sourced there, rather than adding or preserving a generic shell-command resume path for Claude.

After carefully analyzing the original "launch temp" code, i realized this path is too complicated. Actually what needed to restore a session is only a session UUID, instead of creating a temp config file and create a shell script to delete it. So i think simply `exec` the CLI app suffice. 
And in sessions page we can not select providers at all. Launching with currently selected config (in providers page) is the expected behavior.  

The main problem in #222 is the current code launch a new terminal window, which is intuitive for an app already running in a terminal. This problem makes session resume hard to use on headless machines.  This PR fixes this.

SaladDay:
@feiyehua Thanks, I agree with the UX part: restoring in the current terminal is the right direction, especially for headless/SSH use.

The part I still cannot merge is the provider semantics. `claude --resume <id>` is enough for Claude to resume the session, but it does not guarantee the resumed process uses the provider currently selected in cc-switch. The temp settings path is for that: it builds cc-switch's effective provider snapshot and launches Claude with it.

So I would like to keep your current-terminal handoff, but route Claude session restore through the existing temp launch path, passing `--resume <session_id>` as native args and using the session project dir as cwd. That should fix the UX issue without adding a second Claude launch path with different config behavior.

PR #262, opened 2026-06-09
fix(proxy): strip cache_control from OpenAI conversion, guard tool_choice, handle truncated streams

### LINKED ISSUE #257: fix(proxy): Responses→Chat format conversion issues (cache_control, tool_choice, stream truncation)
## Problem

Several Responses→Chat format conversion bugs can cause upstream 400 errors or masked stream failures when routing Codex through Chat Completions providers.

### 1. `cache_control` passthrough causes 400 errors

During Anthropic→OpenAI format conversion, `cache_control` fields are preserved on system messages, text content blocks, and tools. Strict OpenAI-compatible endpoints (e.g. MiniMax, some Chinese providers) reject these unknown fields with 400 errors.

Additionally, when a user message has exactly one text content block with `cache_control`, the content is kept in array-of-objects format instead of being simplified to a plain string. Some providers expect the simpler format.

**Location:** `src-tauri/src/proxy/providers/transform.rs` — system message construction (line ~122), text block construction (line ~339), tool construction (line ~187), and single-block simplification (line ~407).

### 2. `tool_choice` sent without `tools`

In the Responses→Chat conversion, `tool_choice` is copied unconditionally from the input body, but `tools` is only included when non-empty. If all tools are filtered out during conversion, the request contains `tool_choice` without a corresponding `tools` field, which causes errors on providers that enforce this constraint.

**Location:** `src-tauri/src/proxy/providers/transform_codex_chat.rs` — lines 308-315. The `tool_choice` block should be inside the `if !tools.is_empty()` guard.

### 3. Truncated chat streams masked as normal completion

When a Codex Chat→Responses stream ends without receiving `[DONE]` or a `finish_reason`, the code unconditionally calls `finalize()` which emits `response.completed`. This masks truncation as successful completion. The fix should distinguish three cases:
- **Normal completion:** `finish_reason` or `[DONE]` received → `response.completed`
- **Incomplete with output:** substantive output exists but no `finish_reason` → `response.incomplete`
- **Empty truncation:** no output and no `finish_reason` → `response.failed` with `stream_truncated`

**Location:** `src-tauri/src/proxy/providers/streaming_codex_chat.rs` — `finalize()` method (line ~525) and `response_status_from_finish_reason()` in `transform_codex_chat.rs` (line ~1566).

## Expected Behavior

- `cache_control` fields should be stripped from the OpenAI-format request body to avoid 400 errors on strict endpoints
- Single text content blocks should always be simplified to plain string format
- `tool_choice` should be omitted when `tools` is empty
- Truncated streams should be reported as incomplete or failed, not completed

thedavidweng:
## Summary

- Strip `cache_control` fields from Anthropic→OpenAI format conversion (system messages, text blocks, tools) to prevent 400 errors on strict OpenAI-compatible endpoints
- Always simplify single text block content to plain string format
- Move `tool_choice` inside the `tools` guard so it is omitted when tools is empty
- Handle truncated chat streams with a three-way guard: complete normally when `finish_reason` is present, emit incomplete response when output exists without `finish_reason`, emit failed event for empty truncation

### `src-tauri/src/proxy/providers/transform.rs`
- Remove `cache_control` passthrough from system message construction, text block construction, and tool construction
- Remove `cache_control` inheritance logic from `normalize_openai_system_messages`
- Simplify single text block content to always use plain string format

### `src-tauri/src/proxy/providers/transform_codex_chat.rs`
- Wrap `tool_choice` assignment inside the `if !tools.is_empty()` guard

### `src-tauri/src/proxy/providers/streaming_codex_chat.rs`
- Add three-way guard in `finalize()` for truncated streams: when `finish_reason` is absent, check if output exists and emit either `response.completed` with `incomplete` status or `response.failed` with `stream_truncated` error type

Closes #257

### DIFF
--- a/src-tauri/src/proxy/providers/streaming_codex_chat.rs
+++ b/src-tauri/src/proxy/providers/streaming_codex_chat.rs
@@ -533,6 +533,33 @@ impl ChatToResponsesState {
         events.extend(self.finalize_text());
         events.extend(self.finalize_tools());
 
+        // When finish_reason is absent, the stream was truncated without a proper
+        // completion signal.  Distinguish between streams that produced substantive
+        // output (report as incomplete) and those that produced nothing (report as
+        // failed with stream_truncated).
+        if self.finish_reason.is_none() {
+            let output = self.completed_output_items();
+            if output.is_empty() {
+                events.push(self.failed_event(
+                    "Stream truncated before any output was produced".to_string(),
+                    Some("stream_truncated".to_string()),
+                ));
+            } else {
+                let mut response = self.base_response("incomplete", output);
+                response["incomplete_details"] =
+                    json!({ "reason": "stream_truncated" });
+                events.push(sse_event(
+                    "response.completed",
+                    json!({
+                        "type": "response.completed",
+                        "response": response
+                    }),
+                ));
+            }
+            self.completed = true;
+            return events;
+        }
+
         let status = response_status_from_finish_reason(self.finish_reason.as_deref());
         let mut response = self.base_response(status, self.completed_output_items());
         if status == "incomplete" {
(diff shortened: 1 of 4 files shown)


SaladDay:
LGTM, THANKS

PR #263, opened 2026-06-09
fix(codex): generalize cross-turn reasoning cache to all tool call types

### LINKED ISSUE #258: fix(codex): cross-turn reasoning cache missing for custom_tool_call and tool_search_call
## Problem

The cross-turn reasoning cache in Codex Chat history only handles `function_call` type items. When Codex uses `custom_tool_call` (e.g. apply_patch) or `tool_search_call`, their `reasoning_content` is not cached and cannot be restored in follow-up turns.

This causes upstream providers like Kimi/Moonshot and DeepSeek to reject requests with errors like `reasoning_content is missing in assistant tool call message`, because the proxy cannot reconstruct the reasoning context that was present in the original tool call.

## Current Behavior

In `src-tauri/src/proxy/providers/codex_chat_history.rs`:

- `cached_function_call()` (line ~439) only matches `type == "function_call"` and returns `None` for all other types
- `record_response()` (line ~57) uses `.filter_map(cached_function_call)`, so only `function_call` items are recorded
- `enrich_request()` (line ~110) only matches `"function_call"` and `"function_call_output"`, skipping `"custom_tool_call"`, `"tool_search_call"`, and their `*_output` counterparts

## Expected Behavior

The reasoning cache should generalize to the full tool-call triad:
- `function_call` / `function_call_output` (already handled)
- `custom_tool_call` / `custom_tool_call_output` (missing)
- `tool_search_call` / `tool_search_output` (missing)

This ensures that when Codex sends a follow-up request with `previous_response_id` plus a `custom_tool_call_output`, the missing `custom_tool_call` with its `reasoning_content` is restored from the cache, matching the existing `function_call` behavior.

## References

The streaming layer (`streaming_codex_chat.rs`) and transform layer (`transform_codex_chat.rs`) already handle all three tool types correctly. The gap is specifically in the cross-turn history cache.

thedavidweng:
## Summary

The cross-turn reasoning cache in Codex Chat history only handled `function_call` items. When Codex uses `custom_tool_call` (e.g. apply_patch) or `tool_search_call`, their `reasoning_content` was not cached and could not be restored in follow-up turns, causing upstream providers like Kimi/Moonshot and DeepSeek to reject requests with missing reasoning context.

### `src-tauri/src/proxy/providers/codex_chat_history.rs`

- Rename `cached_function_call` to `cached_tool_call` and extend the type check to match `function_call`, `custom_tool_call`, and `tool_search_call`
- Add `is_tool_call_type` and `is_tool_output_type` helper predicates for the full tool-call triad
- Update `enrich_request` to match all tool output types (`function_call_output`, `custom_tool_call_output`, `tool_search_output`) alongside their call counterparts for cross-turn reasoning restoration

The streaming layer and transform layer already handle all three tool types correctly — this change closes the gap specifically in the cross-turn history cache.

Closes #258

### DIFF
--- a/src-tauri/src/proxy/providers/codex_chat_history.rs
+++ b/src-tauri/src/proxy/providers/codex_chat_history.rs
@@ -60,7 +60,7 @@ impl CodexChatHistoryStore {
             .map(|items| {
                 items
                     .iter()
-                    .filter_map(cached_function_call)
+                    .filter_map(cached_tool_call)
                     .collect::<Vec<_>>()
             })
             .unwrap_or_default();
@@ -74,7 +74,7 @@ impl CodexChatHistoryStore {
     }
 
     async fn record_function_call(&self, response_id: Option<&str>, item: &Value) -> bool {
-        let Some(call) = cached_function_call(item) else {
+        let Some(call) = cached_tool_call(item) else {
             return false;
         };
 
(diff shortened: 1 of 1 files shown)


SaladDay:
LGTM, THANKS!!

SaladDay (approved):
LGTM, THANKS!!

PR #264, opened 2026-06-09
fix(usage): correct inflated input_tokens and Zhipu quota routing

### LINKED ISSUE #259: fix(usage): inflated input_tokens in Claude stream parsing and Zhipu quota issues
### 1. Inflated `input_tokens` in Claude stream parsing

Some Anthropic-compatible SSE providers (e.g. Qwen, MiniMax) report the full context (fresh + cached) as `input_tokens` in `message_start`, double-counting the cached portion that is also reported in `cache_read_input_tokens`. This inflates the cacheable-input denominator and pushes the displayed cache hit rate artificially low.

In `src-tauri/src/proxy/usage/parser.rs`, the `from_claude_stream_events` function (line ~288) sets `input_tokens` unconditionally from `message_start`. The `message_delta` handler (line ~336) only reads `input_tokens` when the existing value is exactly zero (`if usage.input_tokens == 0`), so the delta value is silently discarded when `message_start` already provided a (larger) value.

**Fix:** When `message_delta` carries a smaller positive `input_tokens`, prefer it over the `message_start` value and adopt the cache counts from the same usage block.

### 2. Zhipu quota query hardcoded to international endpoint

`query_zhipu()` in `src-tauri/src/services/coding_plan.rs` (line ~184) is hardcoded to `https://api.z.ai`. Users who configured the mainland China preset (`open.bigmodel.cn`) cannot retrieve usage because the international endpoint may be unreachable from their network.

The `detect_provider()` function correctly distinguishes `ZhipuCn` from `ZhipuEn`, but the dispatch (line ~445) collapses both variants into the same `query_zhipu(api_key)` call that ignores the base URL.

**Fix:** Accept `base_url` in `query_zhipu` and route to the correct endpoint based on the configured provider URL.

### 3. Zhipu tier sorting when `nextResetTime` is absent

When the 5-hour bucket is at 0% utilization, Zhipu API omits `nextResetTime`. The tiers are not sorted after construction, and missing `nextResetTime` is not mapped to the correct time bucket, which can cause the weekly bucket to incorrectly claim the five-hour slot in tray and usage quota display.

**Location:** `src-tauri/src/services/coding_plan.rs` — `query_zhipu` function (lines ~240-284).

## Expected Behavior

- Claude stream parsing should prefer the delta `input_tokens` when it is smaller than the start value, avoiding double-counting
- Zhipu quota queries should route to the endpoint matching the user configured `base_url`
- Zhipu tiers should be sorted correctly, with missing `nextResetTime` mapping to the five-hour bucket

thedavidweng:
### 1. Inflated input_tokens in Claude stream parsing

Some Anthropic-compatible SSE providers (e.g. Qwen, MiniMax) report the full context (fresh + cached) as `input_tokens` in `message_start`, double-counting the cached portion. The `message_delta` handler only adopted `input_tokens` when the start value was exactly zero, so the correct (smaller) delta value was silently discarded.

**Fix:** When `message_delta` provides a positive `input_tokens` smaller than the current value, prefer it and sync cache counts from the same delta block.

### 2. Zhipu quota query hardcoded to international endpoint

`query_zhipu()` always hit `https://api.z.ai`, making quota queries fail for users on the mainland China preset (`open.bigmodel.cn`).

**Fix:** Route by the configured `base_url` — `open.bigmodel.cn` users query `open.bigmodel.cn`, others query `api.z.ai`.

### 3. Zhipu tier sorting

When the 5-hour bucket is at 0% utilization, the API omits `nextResetTime`. Without sorting, the weekly bucket could incorrectly claim the five-hour slot.

**Fix:** Sort tiers with missing `nextResetTime` first (just-reset 5-hour bucket), then by reset time ascending.

## Changes

- `src-tauri/src/proxy/usage/parser.rs` — update `from_claude_stream_events` delta handling
- `src-tauri/src/services/coding_plan.rs` — add `zhipu_quota_base` helper, update `query_zhipu` signature, add tier sorting

Closes #259

### DIFF
--- a/src-tauri/src/proxy/usage/parser.rs
+++ b/src-tauri/src/proxy/usage/parser.rs
@@ -333,16 +333,42 @@ impl TokenUsage {
                             {
                                 usage.output_tokens = output as u32;
                             }
-                            // OpenRouter 转换后的流式响应：input_tokens 也在 message_delta 中
-                            // 如果 message_start 中没有 input_tokens，则从 message_delta 获取
-                            if usage.input_tokens == 0 {
-                                if let Some(input) =
-                                    delta_usage.get("input_tokens").and_then(|v| v.as_u64())
+                            // 部分 Anthropic 兼容上游（如 Qwen、MiniMax）在 message_start
+                            // 中将 fresh+cached 合并报告为 input_tokens，导致虚高。
+                            // 当 message_delta 提供了更小的正值 input_tokens 时，优先采用
+                            // delta 的值，并同步更新缓存计数以避免重复计算。
+                            if let Some(input) =
+                                delta_usage.get("input_tokens").and_then(|v| v.as_u64())
+                            {
+                                let delta_input = input as u32;
+                                if delta_input > 0
+                                    && (usage.input_tokens == 0
+                                        || delta_input < usage.input_tokens)
                                 {
-                                    usage.input_tokens = input as u32;
+                                    usage.input_tokens = delta_input;
+                                    // 同步采用 delta 中的缓存计数
+                                    if let Some(cache_read) = delta_usage
+                                        .get("cache_read_input_tokens")
+                                        .and_then(|v| v.as_u64())
+                                    {
+                                        usage.cache_read_tokens = cache_read as u32;
+                                    }
+                                    if let Some(cache_creation) = delta_usage
+                                        .get("cache_creation_input_tokens")
+                                        .and_then(|v| v.as_u64())
+                                    {
+                                        usage.cache_creation_tokens = cache_creation as u32;
+                                    }
+                                }
+                            } else {
+                                // OpenRouter 转换后的流式响应：input_tokens 仅在
+                                // message_delta 中且 message_start 未提供时的回退路径
+                                if usage.input_tokens == 0 {
+                                    // (no input_tokens in delta, keep start value)
                                 }
                             }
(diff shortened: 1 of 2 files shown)


SaladDay:
@thedavidweng Thanks for the focused fix. I found one small issue before this is ready: `zhipu_quota_base()` should use the same lowercasing behavior as `detect_provider()`. Right now an uppercase/custom-cased `OPEN.BIGMODEL.CN` URL is detected as Zhipu CN, but the quota request would still route to `api.z.ai`.

Could you also add small tests for the delta-input override case in the usage parser and the Zhipu quota base selection/sorting behavior? After that, this should be in good shape.

SaladDay:
@thedavidweng Thanks for the fix. I compared this with upstream `farion1231/cc-switch` and the direction looks right, especially the Zhipu CN/EN quota endpoint routing.

One thing I think we should align before merging: upstream now parses Zhipu `TOKENS_LIMIT` tiers more explicitly. It distinguishes `five_hour` vs `weekly_limit` using the `unit` field, then only falls back to reset-time ordering when the unit is missing or unknown. This PR currently sorts the tiers, but still names every `TOKENS_LIMIT` entry as `five_hour`, so weekly quota can be displayed incorrectly.

Could you please update the Zhipu tier parsing to match upstream behavior? After that I think this should be good to merge.

PR #266, opened 2026-06-09
fix(deeplink): preserve custom env fields when importing Claude providers

### LINKED ISSUE #261: fix(deeplink): custom env fields lost when importing Claude providers
## Problem

When importing Claude providers through deeplinks (`ccswitch://v1/import?...`), custom environment variables are silently dropped. Only the hardcoded known keys (`ANTHROPIC_AUTH_TOKEN`, `ANTHROPIC_BASE_URL`, and the four model keys) are preserved.

## Root Cause

Two functions in `src-tauri/src/deeplink/provider.rs` cause the loss:

1. **`build_claude_settings`** (line ~226): Constructs a brand-new empty `serde_json::Map` and only inserts known keys. Any user-defined env fields such as `ANTHROPIC_CUSTOM_HEADERS`, custom proxy variables, or other provider-specific settings are never included.

2. **`merge_claude_config`** (line ~473): When an inline config is provided via the deeplink, this function reads the `env` object but only extracts the hardcoded known keys into the `DeepLinkImportRequest`. Custom env fields are discarded before `build_claude_settings` even runs.

## Contrast with CLI/TUI Path

The CLI/TUI provider creation function `build_claude_settings_config_from_prompt` (in `src/cli/commands/provider_input.rs`, line ~3734) handles this correctly by cloning the existing settings as the base and only mutating known keys, so all sibling keys in `env` survive. Tests at lines 1355-1378 confirm that `EXTRA_ENV` is preserved.

## Expected Behavior

The deeplink import path should preserve all existing environment fields from the source config, matching the behavior of the CLI/TUI provider creation flow. Unknown/custom env keys should be carried through untouched.

thedavidweng:
## Summary

When importing Claude providers through deeplinks (`ccswitch://v1/import?...`), custom environment variables were silently dropped. `build_claude_settings` constructed env from scratch with only known keys (`ANTHROPIC_AUTH_TOKEN`, `ANTHROPIC_BASE_URL`, model keys), discarding any user-defined variables like `ANTHROPIC_CUSTOM_HEADERS` or proxy settings.

The CLI/TUI provider creation path already preserved these fields by cloning existing settings — this fix brings the deeplink import path to parity.

### `src-tauri/src/deeplink/mod.rs`
- Add `custom_env: Option<Map<String, Value>>` field to `DeepLinkImportRequest`

### `src-tauri/src/deeplink/provider.rs`
- In `merge_claude_config`: collect non-standard env keys into `custom_env`
- In `build_claude_settings`: merge `custom_env` entries into the output env

Closes #261

### DIFF
--- a/src-tauri/src/deeplink/mod.rs
+++ b/src-tauri/src/deeplink/mod.rs
@@ -48,6 +48,10 @@ pub struct DeepLinkImportRequest {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub opus_model: Option<String>,
 
+    /// 自定义环境变量（从 inline config 的 env 对象中保留的非标准 key）
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub custom_env: Option<serde_json::Map<String, serde_json::Value>>,
+
     #[serde(skip_serializing_if = "Option::is_none")]
     pub content: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
--- a/src-tauri/src/deeplink/parser.rs
+++ b/src-tauri/src/deeplink/parser.rs
@@ -138,5 +138,6 @@ fn parse_provider_deeplink(
             .get("usageAutoInterval")
             .and_then(|v| v.parse::<u64>().ok()),
         openclaw_config: None,
+        custom_env: None,
     })
 }
--- a/src-tauri/src/deeplink/provider.rs
+++ b/src-tauri/src/deeplink/provider.rs
@@ -256,6 +256,13 @@ fn build_claude_settings(request: &DeepLinkImportRequest) -> serde_json::Value {
         );
     }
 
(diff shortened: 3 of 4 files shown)


SaladDay:
@thedavidweng Thanks for the fix, this is a useful parity improvement.

I found one small blocker before merge: current main supports both Claude auth env keys, `ANTHROPIC_AUTH_TOKEN` and `ANTHROPIC_API_KEY`, but the deeplink merge path only reads/excludes `ANTHROPIC_AUTH_TOKEN`. So an inline config that only has `ANTHROPIC_API_KEY` can still fail with “API key is required”, or can end up with both auth env keys if `apiKey` is provided in the URL.

Could you update `merge_claude_config` to also read `ANTHROPIC_API_KEY`, add it to the known key list, and add a small deeplink import test for that case? After that I think this should be good to merge.

PR #270, opened 2026-06-11
fix(tui): align proxy listen port editing with CLI behavior

### LINKED ISSUE #269: Align TUI proxy listen port editing with CLI behavior
# Problem

The CLI allows changing the selected app's proxy listen port while the local proxy is running, as long as that app does not currently have an active proxy worker.

However, the TUI blocks editing the listen port whenever the local proxy is running. This makes the TUI stricter than the CLI and prevents a valid workflow: updating an inactive app's preferred proxy port while other app routes are still running.

# Current Behavior

In the TUI proxy settings:

- `Listen Address` is blocked while the local proxy is running
- `Listen Port` is also blocked while the local proxy is running
- The user must stop the whole proxy before editing the port, even when the current app has no active worker

In the CLI:

- changing `listen-address` requires stopping the proxy
- changing `listen-port` is only blocked when the selected app has an active proxy worker

# Expected Behavior

The TUI should match the CLI behavior:

- keep `Listen Address` blocked while the local proxy is running
- allow editing `Listen Port` when the proxy is running but the current app has no active worker
- block editing `Listen Port` only when the current app's proxy route is active
- show different guidance for address vs port:
  - address: stop the local proxy
  - port: stop this app's proxy route

# Why

The listen address is global runtime state, so blocking it while the proxy is running makes sense.

The listen port is app-specific preferred worker configuration, so it should be editable when the current app is not actively routed through the proxy. This keeps the TUI consistent with the CLI and avoids requiring users to stop unrelated running proxy routes.

paigeman:
# Summary

The TUI proxy settings blocked editing the listen port whenever the local proxy was running.

This was stricter than the CLI behavior. The CLI already allows changing the selected app's preferred proxy listen port while the proxy is running, as long as that app does not currently have an active proxy worker.

This change brings the TUI behavior in line with the CLI:

- listen address remains blocked while the local proxy is running
- listen port is blocked only when the current app has an active proxy route
- listen port can be edited while other app proxy routes are running

The TUI copy was also split so address and port guidance describe the correct action.

# Changes

## `src-tauri/src/cli/tui/data.rs`

- Add `ProxySnapshot::has_active_worker_for(...)` to centralize active-worker checks for an app.
- Reuse the helper from `routes_current_app_through_proxy(...)`.

## `src-tauri/src/cli/tui/app/content_config.rs`

- Allow opening the listen port editor when the proxy is running but the current app has no active worker.
- Keep listen address editing blocked while the proxy is running.
- Show port-specific guidance based on whether the current app has an active worker.

## `src-tauri/src/cli/tui/app/overlay_handlers/dialogs.rs`

- Apply the same active-worker check when submitting the listen port input.
- Keep submit-time protection in place if the app route starts after the input overlay was opened.

## `src-tauri/src/cli/tui/runtime_actions/settings.rs`

- Keep the runtime fallback aligned with the same rule: block port updates only when the selected app has an active worker.
- Use a port-specific toast when that fallback blocks the update.

## `src-tauri/src/cli/tui/ui/config.rs`

- Show the edit key for listen port when the proxy is running but the current app is not actively routed.
- Hide the edit key for listen port when the current app has an active worker.
- Keep the listen address edit key hidden while the proxy is running.

## `src-tauri/src/cli/i18n/texts/config_actions.rs`

- Split the previous generic stop-before-edit copy into address-specific and port-specific messages.
- Update the running-proxy hint so it describes the separate address and port rules.

## `src-tauri/src/cli/tui/ui/tests.rs`

- Cover editing listen port while the proxy is running for other apps.
- Cover blocking listen port editing when the current app has an active worker.
- Cover submit-time blocking when an active worker appears before confirming the input.
- Cover the updated key hint and status guidance in the TUI rendering tests.

Closes #269

### DIFF
--- a/src-tauri/src/cli/i18n.rs
+++ b/src-tauri/src/cli/i18n.rs
@@ -4260,11 +4260,19 @@ pub mod texts {
         }
     }
 
-    pub fn tui_settings_proxy_stop_before_edit_hint() -> &'static str {
+    pub fn tui_settings_proxy_stop_before_edit_hint(current_app_is_active: bool) -> &'static str {
         if is_chinese() {
-            "请先停止本地代理，再修改监听地址或端口"
+            if current_app_is_active {
+                "修改监听地址：需先停止本地代理。修改监听端口：需先停止当前应用的代理路由。改完后重新启动路由生效。"
+            } else {
+                "修改监听地址：需先停止本地代理。监听端口可以修改。改完后重新启动路由生效。"
+            }
         } else {
-            "Stop the local proxy before editing listen address or port"
+            if current_app_is_active {
+                "Listen address: stop the proxy to edit. Listen port: stop this app's route to edit. Restart routing after changes."
+            } else {
+                "Listen address: stop the proxy to edit. Listen port can be edited. Restart routing after changes."
+            }
         }
     }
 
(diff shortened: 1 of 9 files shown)


SaladDay:
LGTM, THANKS

PR #276, opened 2026-06-12
feat: auto-seed common config snippets from live files on first setup…(首次初始化时自动从配置文件填充通用配置片段)

jiawei666:
… (Claude + Codex + Gemini)

Mirrors the desktop cc-switch initialize_common_config_snippets flow:
- On first startup with empty common config, extract shared settings from live config files (permissions, statusLine, hooks, non-auth env, Codex shared TOML tables, Gemini proxy env) into common_config_<app> snippets.
- Non-destructive: default provider retains full original settings.
- Gate: skipped if snippet already exists or was explicitly cleared.
- OpenCode/Hermes/OpenClaw excluded: their common config merge mechanism is a no-op, and extractors don't match the shape of read_live_settings for additive-mode apps.

### DIFF
--- a/src-tauri/src/database/dao/settings.rs
+++ b/src-tauri/src/database/dao/settings.rs
@@ -74,6 +74,40 @@ impl Database {
         }
     }
 
+    /// 通用配置片段“已被用户显式清空”标记的键名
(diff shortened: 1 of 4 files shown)


SaladDay:
LGTM, THANKS

PR #285, opened 2026-06-18
feat: preserve and edit OpenCode provider modalities field

### LINKED ISSUE #241: [feat] opencode 加上modalities支持
不配置这个. opencode 无法识别图片

```
https://opencode.ai/config.json
              "modalities": {
                "type": "object",
                "properties": {
                  "input": {
                    "type": "array",
                    "items": {
                      "type": "string",
                      "enum": [
                        "text",
                        "audio",
                        "image",
                        "video",
                        "pdf"
                      ]
                    }
                  },
                  "output": {
                    "type": "array",
                    "items": {
                      "type": "string",
                      "enum": [
                        "text",
                        "audio",
                        "image",
                        "video",
                        "pdf"
                      ]
                    }
                  }
                },
                "additionalProperties": false
              },
```

mvanhorn:
## Summary

This change models a `modalities` field on `OpenCodeProviderConfig` and preserves user-authored modalities across the live-config write paths, while staying additive and backward compatible.

## Why

OpenCode image-capable providers advertise their capabilities through a `modalities` block (e.g. `input = ["text", "image"]` and related output modalities). Today `OpenCodeProviderConfig` does not model `modalities` and has no provider-level catch-all, so when cc-switch deserializes an OpenCode provider into the typed config and re-serializes it back to `opencode.json`, a user-supplied provider-level `modalities` block is silently dropped. The result is that switching or editing an image-capable provider through cc-switch can strip the very field that makes it image-capable. This was confirmed by the maintainer on the issue, who kept #241 open for a follow-up release.

## Changes

- `src-tauri/src/provider.rs`: add an optional `modalities: Option<Value>` field and a flattened `extra` catch-all to `OpenCodeProviderConfig`, with `Default` updated accordingly. Both use `skip_serializing_if`, so providers without modalities serialize byte-identically to before (no spurious `"modalities": null` or empty `"extra": {}`).
- `src-tauri/src/opencode_config.rs`: split `set_provider` so raw `Value` callers preserve an existing on-disk `modalities` value when the incoming object omits the key (protecting JSON-editor / TUI edits), while the typed `set_typed_provider` path does not preserve — letting an explicit clear through `OpenCodeProviderConfig { modalities: None, .. }` actually remove the field instead of resurrecting a stale value.

## Tests

Added coverage in `src-tauri/tests/opencode_provider.rs` plus module unit tests:

- A provider with `modalities = { input: ["text","image"] }` round-trips through write -> read unchanged.
- A provider without modalities serializes with no `modalities`/`extra` key (backward-compat golden).
- The raw `set_provider` path preserves an existing on-disk modalities block when the incoming provider omits it.
- The typed `set_typed_provider` path clears modalities when the config explicitly omits them.

Verified locally from `src-tauri/`: `cargo fmt --check`, `cargo clippy --tests`, and the OpenCode/provider tests all pass.

Closes #241

### DIFF
--- a/src-tauri/src/opencode_config.rs
+++ b/src-tauri/src/opencode_config.rs
@@ -63,6 +63,14 @@ pub fn get_providers() -> Result<Map<String, Value>, AppError> {
 }
 
 pub fn set_provider(id: &str, provider: Value) -> Result<(), AppError> {
+    set_provider_value(id, provider, true)
+}
+
+fn set_provider_value(
+    id: &str,
+    mut provider: Value,
+    preserve_modalities: bool,
+) -> Result<(), AppError> {
     let mut full_config = read_opencode_config()?;
 
     if full_config.get("provider").is_none() {
@@ -73,6 +81,21 @@ pub fn set_provider(id: &str, provider: Value) -> Result<(), AppError> {
         .get_mut("provider")
         .and_then(Value::as_object_mut)
     {
+        if preserve_modalities {
+            if let Some(incoming) = provider.as_object_mut() {
+                if !incoming.contains_key("modalities") {
+                    if let Some(modalities) = providers
+                        .get(id)
+                        .and_then(Value::as_object)
+                        .and_then(|existing| existing.get("modalities"))
+                        .cloned()
+                    {
+                        incoming.insert("modalities".to_string(), modalities);
+                    }
+                }
+            }
+        }
+
         providers.insert(id.to_string(), provider);
     }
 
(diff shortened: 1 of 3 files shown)


SaladDay:
LGTM, THANKS

mvanhorn:
Really appreciate the review, @SaladDay. Modeling and preserving the modalities field on OpenCodeProviderConfig keeps user edits intact.

mvanhorn:
Thanks @SaladDay - preserving and editing the OpenCode modalities field keeps user-authored config from getting dropped on write.

PR #295, opened 2026-06-22
fix(proxy): refresh takeover backup on hot-switch

### LINKED ISSUE #292: Claude Code Local configuration does NOT correctly recover
# Issue Description

Local configuration doesn't correctly recover to the selected provider after disable Proxy. The worse thing is that it directly overwrite the selected provider without any explicit reminder.

# Reproduce stage

1. select a provider A with proxy on
2. select a different provider B during proxy on
3. disable proxy
4. select provider C 

then you will find that provider B and provider C configuration will be auto-overwrite by Provider A. 

# Expectation

cc-switch should recover to the selected provider configuration during proxy on/off

unive3sal:
Refresh the restore backup from the selected provider during takeover hot-switches so disabling proxy restores the provider the user chose instead of the one active when takeover began.

Fixes #292

### DIFF
--- a/src-tauri/src/cli/i18n.rs
+++ b/src-tauri/src/cli/i18n.rs
@@ -4270,12 +4270,10 @@ pub mod texts {
             } else {
                 "修改监听地址：需先停止本地代理。监听端口可以修改。改完后重新启动路由生效。"
             }
+        } else if current_app_is_active {
+            "Listen address: stop the proxy to edit. Listen port: stop this app's route to edit. Restart routing after changes."
         } else {
-            if current_app_is_active {
-                "Listen address: stop the proxy to edit. Listen port: stop this app's route to edit. Restart routing after changes."
-            } else {
-                "Listen address: stop the proxy to edit. Listen port can be edited. Restart routing after changes."
-            }
+            "Listen address: stop the proxy to edit. Listen port can be edited. Restart routing after changes."
         }
     }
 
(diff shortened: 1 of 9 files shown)


SaladDay:
@unive3sal Thanks for the fix. The direction makes sense, but I think this still needs one more pass before merge.

The new backup refresh currently uses a two-way merge from the existing restore backup into the selected provider snapshot. That updates keys that exist in the selected provider, but it does not remove provider-owned keys that only existed on the previous provider.

A concrete case:

- provider A restore backup has `env.ANTHROPIC_API_KEY = a-key`, `env.ANTHROPIC_BASE_URL = https://a.example`, and `env.LOCAL_ONLY = kept`
- provider B only has `env.ANTHROPIC_API_KEY = b-key`
- hot-switch to provider B

After `update_live_backup_from_provider`, the restore backup becomes:

```json
{
  "env": {
    "ANTHROPIC_API_KEY": "b-key",
    "ANTHROPIC_BASE_URL": "https://a.example",
    "LOCAL_ONLY": "kept"
  }
}
```

So disabling proxy would restore provider B's key mixed with provider A's base URL. That still leaves part of #292 unfixed.

I think this needs a base-aware merge: base = previous selected provider snapshot, local = existing restore backup, incoming = newly selected provider snapshot. That would preserve local-only fields like `LOCAL_ONLY`, while removing old provider-owned fields that the new provider no longer has. Codex TOML likely needs the same stale-key consideration too, not only Claude JSON.

Could you add a regression test for this case? The current tests cover overwriting shared keys and preserving local-only keys, but not removing stale provider-owned keys.

SaladDay:
LGTM, THANKS

PR #317, opened 2026-06-29
feat: resolve providers by name in use/provider switch

### LINKED ISSUE #301: 【FR】：cc-switch use <id> 支持 name
当前 `cc-switch use` 和 `cc-switch provider switch` 命令只支持 **id**，能否兼容一下 **name**，不然 id 实在太长记不住

mvanhorn:
## Summary
`cc-switch use <name>` and `cc-switch provider switch <name>` now accept a provider's human-readable name, not just its id. An exact id match still takes precedence, so existing id-based usage is unchanged.

## Why this matters
Issue #301 reports that provider ids are long UUID-like strings users cannot remember, so the id-only lookup in `switch_provider` (src-tauri/src/cli/commands/provider.rs) was hard to use. Both `use` (dispatched in main.rs as `ProviderCommand::Switch`) and `provider switch` funnel through `switch_provider`, so resolving the argument once there covers both commands.

## What changed
When the exact-id lookup misses, `switch_provider` now falls back to a case- and whitespace-insensitive match on the provider `name`. A single name match resolves to that provider's real id, which is then used for the `ProviderService::switch` call, plugin sync, and success message. If two providers share a name, it returns an error listing the candidate ids so the user can disambiguate by id. The original "Provider '<arg>' not found" wording is preserved when nothing matches.

## Testing
Added four integration tests in src-tauri/tests/provider_commands.rs covering: switch by name (case/whitespace-insensitive), exact-id precedence over a colliding name, unknown id/name not-found, and the ambiguity error listing candidate ids. Ran `cargo fmt --check`, `cargo clippy --tests`, and `cargo test --test provider_commands provider_switch` plus the provider unit tests; all pass.

Fixes #301

### DIFF
--- a/src-tauri/src/cli/commands/provider.rs
+++ b/src-tauri/src/cli/commands/provider.rs
@@ -17,6 +17,7 @@ use crate::error::AppError;
 use crate::provider::{AuthBinding, AuthBindingSource, ClaudeApiKeyField, Provider, ProviderMeta};
 use crate::services::{AuthService, ManagedAuthAccount, ProviderService};
 use crate::store::AppState;
+use indexmap::IndexMap;
 use inquire::{Confirm, Select};
 
 const AUTH_PROVIDER_CODEX_OAUTH: &str = "codex_oauth";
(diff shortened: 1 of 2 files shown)


SaladDay:
LGTM, THANKS !!

PR #319, opened 2026-06-30
fix(tui): prefer ansi256 for bare TERM=xterm without truecolor signal

### LINKED ISSUE #60: 5.x 版本 UI 展示异常
Image

``` shell
root@ubuntu:~# cc-switch --version
cc-switch 5.0.1
root@ubuntu:~# lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.6 LTS
Release:	18.04
Codename:	bionic
```

感谢大佬的开源项目，之前是 4.x 版本的，使用 `cc-switch update` 升级 5.0.1 后出现这个问题，试了一下无法降级：`Error: 数据库错误: 数据库版本过新（6），当前应用仅支持 5，请升级应用后再尝试。`，只能删除重装处理，不知道是不是个例。

mvanhorn:
## What

Make bare `TERM=xterm` (with no truecolor signal) prefer `ColorMode::Ansi256` instead of falling through to `TrueColor`, regardless of whether the session is over SSH.

## Why

Issue #60 reports garbled colors in the TUI on Xshell 8. The decisive diagnostic from the reporter was `TERM=xterm`, empty `COLORTERM`, no `TMUX`, no `SSH_TTY`, running directly in Xshell, with `CC_SWITCH_COLOR_MODE=ansi256 cc-switch` confirmed as the working fix.

In `detected_color_mode()` that environment fell through every branch (`known_ansi256_terminal`, the `COLORTERM`/`TERM` truecolor checks, and `ssh_plain_xterm_prefers_ansi256` which also required an `SSH_TTY`) and hit the final `ColorMode::TrueColor` fallback. Emitting 24-bit escape sequences to a terminal that doesn't reliably support them produces the wrong colors. A bare `TERM=xterm` (exactly `xterm`, not `xterm-256color`) with no `COLORTERM` is a strong signal the terminal is not truecolor-capable, and the existing SSH-only gate was too narrow since the same terminal misbehaves locally.

## Change

- Generalize `ssh_plain_xterm_prefers_ansi256` into `plain_xterm_prefers_ansi256`: it now matches an exact `TERM=xterm` (case-insensitive) and no longer requires an SSH session. The `COLORTERM`/`TERM` truecolor checks still run first in `detected_color_mode()`, so reaching this branch already means no truecolor signal was advertised.
- `xterm-256color`, explicit `COLORTERM=truecolor`/`24bit`, and the `CC_SWITCH_COLOR_MODE` override all stay on the truecolor path, so genuinely-capable terminals are unaffected. An unset `TERM` also still falls through to truecolor.

Tradeoff: a small number of terminals advertise `TERM=xterm` yet do support truecolor. This change trades that rare case for correct rendering on the common plain-`xterm` case; those users can still force truecolor with `CC_SWITCH_COLOR_MODE=truecolor`.

## Testing

Tests use the existing `EnvGuard` harness in `theme.rs`:

- The previous `theme_keeps_truecolor_for_plain_xterm_without_ssh_signal` test (which asserted the reported-buggy behavior) is updated to `theme_uses_ansi256_for_plain_xterm_without_truecolor_signal` and now asserts `Ansi256`.
- Added `explicit_truecolor_override_beats_plain_xterm_auto_fallback`: `CC_SWITCH_COLOR_MODE=truecolor` with `TERM=xterm` still returns `TrueColor`.
- `xterm-256color` + `COLORTERM=truecolor`, unset `TERM`, the SSH plain-xterm case, and `NO_COLOR`/`none` remain covered and pass.

```
cd src-tauri && cargo test --lib cli::tui::theme
# 17 passed; 0 failed
```

`cargo fmt --check` and `cargo clippy --lib` are clean for the touched file.

Closes #60

### DIFF
--- a/src-tauri/src/cli/tui/theme.rs
+++ b/src-tauri/src/cli/tui/theme.rs
@@ -75,11 +75,10 @@ fn known_ansi256_terminal() -> bool {
         .unwrap_or(false)
 }
 
-fn ssh_plain_xterm_prefers_ansi256() -> bool {
-    std::env::var_os("SSH_TTY").is_some()
-        && std::env::var("TERM")
-            .map(|value| value.eq_ignore_ascii_case("xterm"))
-            .unwrap_or(false)
+fn plain_xterm_prefers_ansi256() -> bool {
+    std::env::var("TERM")
+        .map(|value| value.eq_ignore_ascii_case("xterm"))
+        .unwrap_or(false)
 }
 
 fn detected_color_mode() -> ColorMode {
@@ -99,7 +98,7 @@ fn detected_color_mode() -> ColorMode {
         return ColorMode::TrueColor;
     }
 
-    if ssh_plain_xterm_prefers_ansi256() {
+    if plain_xterm_prefers_ansi256() {
         return ColorMode::Ansi256;
     }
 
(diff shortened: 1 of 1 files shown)


SaladDay:
LGTM, THANKS!!

PR #320, opened 2026-06-30
fix(proxy): emit spec-complete message_start (content/stop_reason/stop_sequence)

lishuceo:
## Problem

When the proxy converts non-Anthropic backends (OpenAI-compatible chat, Gemini, OpenAI Responses) into an Anthropic `/v1/messages` SSE stream, the `message_start` event's `message` snapshot omitted `content`, `stop_reason`, and `stop_sequence`.

The Anthropic streaming spec requires that snapshot to be a full Message with `content: []`. The official `@anthropic-ai/sdk` stream accumulator runs `snapshot.content.push(...)` and throws `Cannot read properties of undefined (reading 'push')` when `content` is missing — forcing clients into per-turn non-streaming fallback (double billing) or intermittently surfacing as a silent empty end-of-turn that terminates agent sessions mid-task.

## Fix

Add the three spec-required fields (`content: []`, `stop_reason: null`, `stop_sequence: null`) to **every** `message_start` construction site across the three transform paths:

- `streaming.rs` — Anthropic / OpenAI-compatible chat (1 site)
- `streaming_gemini.rs` — Gemini → Anthropic (2 sites)
- `streaming_responses.rs` — OpenAI Responses → Anthropic (3 sites, including the tool-call fallback path)

Verified field-for-field against the documented `message_start` shape in the Anthropic streaming docs.

## Tests

Adds a regression test per transform path asserting `message_start` carries the spec-required fields (including the Responses tool-only path).

- `cargo fmt --check` clean
- `cargo clippy` clean (no new warnings in the changed files)
- Streaming unit tests + `proxy_claude_streaming` / `proxy_claude_response_parity` / `proxy_claude_openai_chat` integration targets pass

### DIFF
--- a/src-tauri/src/proxy/providers/streaming.rs
+++ b/src-tauri/src/proxy/providers/streaming.rs
@@ -166,6 +166,15 @@ pub fn create_anthropic_sse_stream(
                                         "id": message_id.clone().unwrap_or_default(),
                                         "type": "message",
                                         "role": "assistant",
+                                        // Anthropic SSE spec requires these fields in the
+                                        // message snapshot; the official SDK's stream
+                                        // accumulator does `snapshot.content.push(...)` and
+                                        // crashes ("Cannot read properties of undefined")
+                                        // when `content` is missing, forcing clients into
+                                        // non-streaming fallback (or silent stream loss).
+                                        "content": [],
+                                        "stop_reason": serde_json::Value::Null,
+                                        "stop_sequence": serde_json::Value::Null,
                                         "model": current_model.clone().unwrap_or_default(),
                                         "usage": start_usage
                                     }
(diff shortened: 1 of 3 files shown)


lishuceo:
@SaladDay Friendly ping 🙏 This fixes a spec-compliance bug where the proxy's message_start omits content/stop_reason/stop_sequence, which crashes the official Anthropic SDK stream accumulator and forces clients into non-streaming fallback. Covered by regression tests across all three transform paths; fmt/clippy/streaming tests pass. Happy to adjust anything you'd like.

SaladDay:
@lishuceo I will review it in a while. Thank you for your contribution.

SaladDay:
LGTM, THANKS!!!

PR #337, opened 2026-07-10
Add provider launch command

felinae98:
Adds `provider launch` as a CLI entry point for temporary Claude/Codex agent launches without switching the current provider.\n\nValidation:\n- `cargo fmt --check`\n- `cargo test provider_launch`\n- `cargo test start_codex_launches_resolved_provider`\n- `cargo test start_claude_launches_resolved_provider`

### DIFF
--- a/src-tauri/src/cli/commands/provider.rs
+++ b/src-tauri/src/cli/commands/provider.rs
@@ -1,7 +1,7 @@
 use clap::{Subcommand, ValueEnum};
-use std::{collections::HashSet, path::PathBuf};
+use std::{collections::HashSet, ffi::OsString, path::PathBuf};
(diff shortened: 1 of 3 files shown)


SaladDay:
Thanks for the PR. We already support the same workflow with:

```bash
cc-switch start claude <provider>
cc-switch start codex <provider>
```

This adds a second command path over the same implementation without adding new behavior, so I don't think we should maintain both.

If you prefer a shorter command locally, a shell alias should cover it:

```bash
alias cc-claude='cc-switch start claude'
alias cc-codex='cc-switch start codex'
```

I'm going to close this one, but thanks for taking the time to put it together.

PR #339, opened 2026-07-10
fix: refresh Claude live config after common updates during takeover

### LINKED ISSUE #338: Claude common config updates do not refresh active live config during takeover
## Summary
When Claude proxy takeover is already active, updating Claude common config refreshes the DB snippet but does not refresh the active live config (`~/.claude/settings.json`).

In this state, `cc-switch` reports success / refreshed live config, but snippet-owned fields such as `statusLine` or simple env keys never appear in the active Claude live settings.

## Scope
This report is specifically about the Claude takeover-active path.

I am **not** claiming that common config is generally broken:
- when proxy takeover is inactive, `config common set` updates Claude live config as expected
- when common config is set before `proxy enable`, the resulting taken-over live config also keeps the common fields
- the stale behavior appears when takeover is already active and then common config is updated

## Reproduction
Minimal repro uses a simple env key instead of `statusLine`, so no helper script is required.

1. Start with a Claude provider whose `commonConfigEnabled=true`
2. Enable Claude proxy takeover
3. Run a common-config update, for example a snippet that adds:
   - `env.MY_TEST_KEY = "1"`
4. Inspect both:
   - the stored common config in cc-switch
   - the active Claude live config at `~/.claude/settings.json`

## Expected behavior
Both should reflect the new common-config fields.

## Actual behavior
- cc-switch stores the updated Claude common config
- but the active Claude live config remains stale
- takeover-only fields (proxy URL / managed placeholder) stay in place, but newly added common fields do not appear

In my real-world case this showed up as `statusLine` existing in cc-switch common config while Claude Code never displayed it, because `~/.claude/settings.json` never received the field.

## Notes from local investigation
The likely problematic path seems to be the post-commit flow used when takeover is already active:
- common-config update builds a post-commit action with `takeover_active=true`
- this goes through the `ProxyLiveBackup` effect
- backup/snapshot handling runs, but the active Claude live config is not refreshed through the same effective-settings path used elsewhere

I have a local patch and regression test prepared and can open a PR referencing this issue.

mydelren:
## Summary
- refresh active Claude live settings after common-config updates while takeover is already active
- build the takeover backup snapshot from the in-flight common-config snippet instead of reloading the old stored snippet
- add a regression test covering common-config updates during active Claude takeover

## Problem
When Claude takeover is already active, updating Claude common config can persist the snippet in cc-switch while leaving the active live config stale.

In practice this means fields like `statusLine` (or simpler env keys) can exist in Claude common config but never appear in `~/.claude/settings.json` until some later operation rewrites live config.

## Root cause addressed here
The takeover-active post-commit path used `ProxyLiveBackup` and saved a refreshed backup snapshot, but it did not refresh the active Claude live config. It also rebuilt backup state from the stored common snippet instead of the in-flight snippet being applied by the current operation.

This patch:
1. passes the current common-config snippet through the takeover backup snapshot path
2. refreshes active Claude live config after saving the takeover backup snapshot

## Verification
Already verified locally before preparing this PR:
- `cargo build --lib`
- targeted integration test for the new regression case
- takeover + common_config related test selection
- full lib test suite
- full `provider_commands` integration tests

## Regression coverage
Added:
- `updating_claude_common_config_under_takeover_refreshes_active_live_settings`

## Issue
Closes #338

### DIFF
--- a/src-tauri/src/services/provider/mod.rs
+++ b/src-tauri/src/services/provider/mod.rs
@@ -776,6 +776,7 @@ impl ProviderService {
                     action.app_type.as_str(),
                     &action.provider,
                     action.previous_provider.as_ref(),
+                    action.common_config_snippet.as_deref(),
                 ))
                 .map_err(AppError::Message)?;
             PreparedPostCommitEffect::ProxyLiveBackup(backup_snapshot)
@@ -822,6 +823,16 @@ impl ProviderService {
                         .save_live_backup_snapshot(prepared.action.app_type.as_str(), snapshot),
                 )
                 .map_err(AppError::Message)?;
+
+                // Refresh active live settings under takeover
+                if matches!(prepared.action.app_type, AppType::Claude) {
+                    futures::executor::block_on(
+                        state.proxy_service.sync_claude_live_from_provider_while_proxy_active(
+                            &prepared.action.provider,
+                        ),
+                    )
+                    .map_err(AppError::Message)?;
+                }
             }
         }
 
--- a/src-tauri/src/services/proxy.rs
+++ b/src-tauri/src/services/proxy.rs
@@ -1591,7 +1591,7 @@ impl ProxyService {
         original_live: &Value,
         provider: &Provider,
     ) -> Result<Value, String> {
-        let mut provider_snapshot = self.build_live_snapshot_from_provider(app_type, provider)?;
+        let mut provider_snapshot = self.build_live_snapshot_from_provider(app_type, provider, None)?;
         Self::apply_codex_unified_session_bucket_to_backup(
             app_type,
             provider,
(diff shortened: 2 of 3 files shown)


SaladDay:
Thanks for the PR. The direction is right, but I found two blocking cases:

- Clearing the common config, or replacing it with a smaller snippet, leaves removed keys in the takeover restore backup. Those keys return after takeover is disabled.
- If refreshing the active Claude config fails after the new backup is saved, rollback does not restore the previous backup.

I reproduced the first case with both clear and reduced snippets. Could you address these cases and add tests for clear/reduced snippets, restore after disabling takeover, and live-write failure rollback? `cargo fmt --check` also currently fails on this branch.

PR #341, opened 2026-07-13
fix: prevent empty tool_call delta from overwriting accumulated call_…

### LINKED ISSUE #340: Can not use tool calls when using DsahScope deekseep-v4-pro in Codex
When using DashScope's deepseep-v4-pro with the chat-completion proxy enabled, tool calls cannot be invoked.
The Language Model return unsupported call.

Image

```
› Can you help me to write a bubble sort code


• The user wants me to write a bubble sort implementation. This is a straightforward request. Let me create a clean, well-written
  bubble sort implementation. I'll write it in Python since that's a common language for algorithms and the user hasn't specified a
  language.


  Let me create a file with a bubble sort implementation that includes:


  1. The basic bubble sort function
  2. An optimized version (early exit when no swaps)
  3. A simple demo/example usage


  I'll create it in the current working directory.

• The cat with heredoc approach failed. I should use apply_patch to create the file instead, or use exec_command to write the file.
  Let me use apply_patch to create a new file.

• The tool I'm trying to call doesn't seem to exist. Let me look at the available tools again. I have apply_patch, exec_command,
  write_stdin, update_plan, request_user_input, view_image, get_goal, create_goal, update_goal, tool_search.


  I need to use apply_patch to create a file or exec_command to write it. Let me use apply_patch since the instructions say "Use
  apply_patch for manual code edits. Do not create or edit files with cat or other shell write tricks."


  But I also need to look at how apply_patch works. Let me try using it to create a new file.

• The tool call format here seems to require the name parameter. Looking at the documentation again, the tool I should use is
  apply_patch. Let me try the correct format.


  Looking at the available tools:


  - apply_patch - parameters: input (string)


  Let me try the proper format.

• Hmm, the apply_patch tool seems to be having issues. Let me try using exec_command with a heredoc approach, but the permissions
  instructions said "Do not create or edit files with cat or other shell write tricks." However, maybe I can use tee or printf?
  Actually, let me just try a simple exec_command with python3 -c to write the file, or use tee.

  Actually, wait - let me re-read the instructions. "Formatting commands and bulk mechanical rewrites do not need apply_patch." -
  maybe I can use shell commands for this since it's creating a new file, not editing an existing one.


  Let me try exec_command with a heredoc to cat.

• I'm having issues with the tool call format. Let me look at the examples more carefully. The tools available are:


  - exec_command - cmd, justification, login, max_output_tokens, prefix_rule, sandbox_permissions, shell, tty, workdir,
    yield_time_ms


  - apply_patch - input


  The format should be:


  <｜DSML｜tool_calls>
  <｜DSML｜invoke name="exec_command">
  <｜DSML｜parameter name="cmd" string="true">...</｜DSML｜parameter>
  </｜DSML｜invoke>
  </｜DSML｜tool_calls>


  Wait, I think my previous attempts were missing the name attribute on the <｜DSML｜invoke> tag. Let me check... No, looking at the
  previous calls, I had <｜DSML｜invoke name="exec_command"> in one of them. But looking more carefully, the tool call format seems
  correct. Let me try a simpler command first to verify the tool works.

• The tool calls are failing with "unsupported call". Let me look at the system message again more carefully. Looking at the
  instructions:


  "You have access to a set of tools to help answer the user's question. You can invoke tools by writing a "<｜DSML｜tool_calls>"
  block like the following:"


  Then it shows examples. My format seems correct. But the error is "unsupported call". Let me check if maybe the tool name is wrong
  or if I need to use different names. Let me look at the available tools list again:


  - exec_command
  - write_stdin
  - update_plan
  - request_user_input
  - apply_patch
  - view_image
  - get_goal
  - create_goal
  - update_goal
  - tool_search
  These are the correct names. Let me try a very simple call.

• Every call I make is returning "unsupported call." There might be something wrong with the formatting. Let me re-read the
  instructions carefully:


  "You can invoke tools by writing a "<｜DSML｜tool_calls>" block like the following:"


  And then it shows examples. But the examples aren't shown in the system message. Let me check the exact format... The issue might
  be that I need to match the exact format precisely.


  Looking at the system message more carefully, I see the template for tool calls. Let me try a different format - maybe the name
  attribute on <｜DSML｜invoke> is necessary. All my calls have <｜DSML｜invoke name=""> with an empty name. Let me try with the
  proper name.


  Wait, I was indeed using <｜DSML｜invoke name=""> with empty name! Let me fix that.
```

leaf-bridge:
## Summary

Fix streaming tool call conversion where `call_id` and `name` are overwritten by empty delta chunks.

In `push_tool_call_delta`, both `id`/`call_id` and `name` assignments lacked empty-string protection. When upstream models (e.g. GLM, qwen via DashScope/Aliyun) return empty `id` and `name` in tool_call delta chunks after the first one, the accumulated values are overwritten with empty strings. This causes `response.output_item.done` and `response.completed` to carry empty `call_id` and `name`, leading Codex to reject the tool call with `unsupported call:`.

This PR adds non-empty checks for both `id` and `name` assignments, so that empty deltas are treated as "continue previous value" rather than "clear". Verified locally that streaming tool calls work correctly after the fix.

## Related Issue

Fixes #340

## Screenshots

N/A (no UI changes)

| Before | After |
|--------|-------|
| `call_id` and `name` empty in `output_item.done` | Both fields correctly preserved |

## Checklist

- [x] `cargo clippy` passes
- [x] `cargo test` passes

### DIFF
--- a/src-tauri/src/proxy/providers/streaming_codex_chat.rs
+++ b/src-tauri/src/proxy/providers/streaming_codex_chat.rs
@@ -426,10 +426,14 @@ impl ChatToResponsesState {
         {
             let state = self.tools.entry(chat_index).or_default();
             if let Some(id) = id_delta {
-                state.call_id = id;
+                if !id.is_empty() {
+                    state.call_id = id;
+                }
             }
             if let Some(name) = name_delta {
-                state.name = name;
+                if !name.is_empty() {
+                    state.name = name;
+                }
             }
             if !args_delta.is_empty() {
                 state.arguments.push_str(&args_delta);


SaladDay:
Thanks @lao-der for the clear reproduction and accurate diagnosis. The non-empty guards for `call_id` and `name` are correct.

While this PR was open, the same fix landed on `main` in [`793a6ba4`](https://github.com/SaladDay/cc-switch-cli/commit/793a6ba4dcddb0fcc0f4e2c175ccf1f11a99e867), together with broader handling and regression coverage for late identity fragments. [`91ef1e5b`](https://github.com/SaladDay/cc-switch-cli/commit/91ef1e5bbb435274b9f49f90ca7e58f5f61a3885) also preserves parallel tool-call order. Both fixes shipped in [`v5.9.1`](https://github.com/SaladDay/cc-switch-cli/releases/tag/v5.9.1), and the DashScope empty-continuation regression test passes on the current `main`.

GitHub now reports this PR as conflicting because the affected code has already been replaced, so resolving the conflict would leave no remaining change to merge. I am closing this as superseded rather than asking you to rebase.

Thank you again for investigating the DashScope/GLM/Qwen behavior and contributing the fix.

PR #380, opened 2026-07-27
feat(sessions): add usage analytics

SaladDay:
## What this adds

- Adds `sessions usage --provider claude|codex --range today|7d|30d`, with table and JSON output.
- Extends the TUI Usage screen with Models, Providers, Sessions, and Recent views.
- Imports Claude and Codex transcript usage, attributes proxy rows to sessions, and avoids double-counting imported and proxied usage where identity evidence is available.
- Keeps detailed rows for 31 calendar days while retaining older aggregate usage.
- Adds bounds for session/model result sets and keeps the JSON session ID contract lossless.

The design keeps the existing upstream Sessions and Usage flow as the base, with selected ideas from Tokscale. This PR does not add a dynamic pricing catalog.

## Boundaries

- The persistent SQLite schema is unchanged.
- `Cargo.toml` and `Cargo.lock` are unchanged.
- Claude and Codex are the strict support targets. Other providers keep the upstream best-effort behavior.
- No new cross-provider compatibility framework was added.

## Handoff notes

This is a work-in-progress handoff. A final blind review found the following open items:

1. **P1: ordinary startup can block on a usage import lock.** Startup and periodic maintenance currently take the blocking `.session-usage.lock`. If another process is running a long `sessions sync-usage --all`, unrelated commands can wait indefinitely during `Database::init()`. Normal maintenance should use a non-blocking try-lock and defer when busy. Migration, rebuild, and explicit import paths may continue to block. Relevant code is in `database/mod.rs` and `services/session_usage.rs`.
2. **P1: WebDAV compatibility was not bumped after making `session_log_sync` local-only.** Backup export now omits this table's data and import preserves local cursors, but `DB_COMPAT_VERSION` is still 6. Older clients can accept the snapshot, clear their local cursor table, rescan pruned detail, and permanently double-count usage. Either bump the sync compatibility version or restore compatible export semantics. Relevant code is in `database/backup.rs` and `services/sync_protocol.rs`.
3. **P2: whitespace-only client session IDs can become trusted.** Claude/Codex extraction accepts some whitespace-only IDs and stores a trust marker. Exact transcript backfill then refuses to replace them, so those rows remain absent from Sessions. Reject blank-after-trim IDs and allow marked blank identities to be overwritten. Relevant code is in `proxy/session.rs` and `services/session_identity.rs`.

The two P1 items should be fixed before merge. The P2 item is also worth fixing because Claude and Codex are the strict support targets.

Two low-priority UI behaviors were intentionally left alone: an empty CLI result always suggests running sync, and the table can wrap on extremely narrow terminals.

## Verification completed

- `cargo fmt --all -- --check`
- `cargo check --quiet`
- Library test suite: 3,615 passed
- `provider_commands`: 29 passed
- Focused identity, session query, rollup, rebuild, migration, proxy logging, and TUI tests passed
- `git diff --check`
- Confirmed zero diff in `src-tauri/src/database/schema.rs`, `src-tauri/Cargo.toml`, and `src-tauri/Cargo.lock`

`cargo test --tests` remains blocked by baseline compile errors in unchanged `settings_current_provider.rs` and `settings_visible_apps.rs`; this branch does not modify either file.


SaladDay:
Closing for now — shelving this feature.

The Sessions + Usage work here is ~7.7k lines, but the actual user-visible surface is one new `Sessions` tab plus a `sessions usage` CLI command; Models / Providers / Recent, the metric cards, trend chart and sparkline all already existed upstream. The rest is backend plumbing to make proxy-logged usage and transcript-imported usage agree on what counts once.

That reconciliation turned out to be the wrong shape for the constraints (SQLite schema must stay byte-identical with upstream, so identity had to be re-derived at query time from overloaded columns). It was reworked to upstream's write-time dedup plus the upstream read-side safety net, which did converge — but the cost/benefit against what actually ships to users isn't there yet.

Reopening later would start from the write-time design, not the read-time reconciliation.

PR #382, opened 2026-07-28
fix(pricing): seed MiniMax-M3 in built-in model pricing registry

octo-patch:
Reason: parameter-refresh — MiniMax-M3 was missing from the built-in model pricing registry while MiniMax-M2.7 pricing was already present.

## Changes

- Add the `minimax-m3` row to `seed_model_pricing` in `src-tauri/src/database/schema.rs` with input `$0.6`, output `$2.4`, cache-read `$0.12`, and cache-write `$0` per million tokens (USD), matching the MiniMax-M3 pricing published in the provider parameter set. The MiniMax section already seeded MiniMax-M2.7 with its current pricing; M3 was absent.
- Extend `schema_model_pricing_is_seeded_on_init` in `src-tauri/src/database/tests.rs` to assert the seeded `minimax-m3` row carries the expected input, output, cache-read, and cache-write values, guarding against future regressions.

## Checks

- `cargo fmt --check` (passes)
- `cargo test --lib schema_model_pricing_is_seeded_on_init` (1 passed)


SaladDay:
Thanks for the focused PR. The seed path and regression test look good. I verified that `cargo fmt --check` and the targeted unit test pass, and the branch is mergeable without conflicts.

One pricing value needs to be corrected before merge. MiniMax's current pay-as-you-go page marks the 50% discount as permanent. For requests with up to 512K input tokens, the effective rates are `$0.30` input, `$1.20` output, and `$0.06` cache read per million tokens. The proposed `$0.60 / $2.40 / $0.12` row would overstate standard-tier M3 costs by 2x.

Please update the seed row and test expectation to `0.30 / 1.20 / 0.06 / 0`. Since this fork does not have an existing M3 row yet, no migration or repair entry is needed here.

The same stale values are present upstream, so I reported them in https://github.com/farion1231/cc-switch/issues/5841.

Official pricing: https://platform.minimax.io/subscribe/token-plan?tab=api-enterprise

SaladDay:
Thank you for catching the missing MiniMax-M3 pricing entry and for adding regression coverage. Your contribution prompted us to review and synchronize the broader built-in pricing catalog.

The resulting update has landed on `main` in https://github.com/SaladDay/cc-switch-cli/commit/4bfc982f. It includes MiniMax-M3 at the current effective ≤512K rate, the other upstream catalog updates, and guarded repairs for existing built-in values without overwriting user-customized prices.

This PR is now superseded by that commit, so I am closing it instead of merging the duplicate change. Thanks again for taking the time to contribute.

PR #396, opened 2026-08-02
fix(codex): recover legacy thread names

huzhongyyuan:
Explicit opt-in repair redesign is in commit 7eee7adf. See the latest comment for safety details.

### DIFF
--- a/src-tauri/src/cli/commands/sessions.rs
+++ b/src-tauri/src/cli/commands/sessions.rs
@@ -16,6 +16,12 @@ use crate::session_manager::{self, SessionMessage, SessionMessageBatch, SessionM
 
 #[derive(Subcommand, Debug, Clone)]
 pub enum SessionsCommand {
+    /// Repair legacy Codex session names (dry-run unless --apply is passed)
+    RepairCodexNames {
+        /// Apply the reported repairs to the Codex state database
+        #[arg(long)]
+        apply: bool,
+    },
     /// List saved assistant sessions
     List {
         /// Scan a specific provider instead of --app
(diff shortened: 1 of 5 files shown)


huzhongyyuan:
Thanks for pointing out that automatic startup migration was not appropriate. I redesigned the PR as an explicit opt-in command: `cc-switch sessions repair-codex-names`. It is dry-run by default, and only `--apply` writes changes.\n\nThe command strictly backfills only `history_mode = paginated` rows whose `threads.name` is NULL or empty, using non-empty `thread_name` values from `session_index.jsonl`. Existing non-empty names are never changed or cleared. Archive state and provider fields remain untouched, and the old startup automatic migration has been removed.\n\nThe redesign also removes the repair ledger and all ambiguous ownership/ABA paths. It adds immutable dry-run handling, WAL/SHM/journal protection, bounded config/index/database work, index-plan validation, pre-write SQLite backups, conditional updates, and post-commit identity/source checks.\n\nValidation passed: `cargo check`, `cargo fmt --check`, three focused Codex repair tests, two CLI parser tests, and `git diff --check`. The updated commit is `7eee7adf`. Please review this redesigned explicit-command version again. If you prefer an even smaller PR scope, I am happy to adjust it.

SaladDay:
Thanks for redesigning this as an explicit, dry-run-by-default repair command — this is a much safer direction.

I found two remaining issues that should be addressed before merging:

1. There is still a source-update race between the final `session_index.jsonl` digest check and the SQLite commit. If a newer name is appended in that window, the older name is committed and the post-commit check only reports the problem. A retry cannot repair it because the row is no longer blank.

2. A non-applicable default DB can block a valid configured `sqlite_home` DB. The resolver always includes the default DB, but a stale default DB without `threads.name` or `history_mode` fails preflight and prevents all repairs. I reproduced this with an old-schema default DB and a valid configured DB. Missing repair-related columns should probably make that DB inapplicable/skipped, while genuine open or corruption errors should still fail.

It would also be helpful to add public-entry-point tests covering the multi-DB/`sqlite_home` case and the final source-change window. The focused tests and CLI parser tests otherwise pass, and the `threads.name`/paginated scope looks correct. Thanks again for the thoughtful revision!

PR #421, opened 2026-08-21
feat: make CLI dependencies optional

cindyrui71:
## Summary

- add a default-enabled `cli` Cargo feature and require it for the `cc-switch` binary
- gate CLI/TUI modules and their direct dependencies so library consumers can build with `default-features = false`
- keep shared localization available to core library modules
- add CI coverage for the library-only build, test compilation, and all-target direct dependency boundary

Default builds remain unchanged: the `cli` feature is enabled by default.

## Validation

- `cargo check`
- `cargo test --lib --no-run`
- `cargo test --bin cc-switch --no-run`
- `cargo check --lib --no-default-features`
- `cargo check --lib --no-default-features --features test-hooks`
- `cargo test --lib --no-default-features --no-run`
- `cargo fmt --check`
- `git diff --check`
- parsed `.github/workflows/rust-ci.yml` as YAML
- verified the all-target, no-default direct dependency graph excludes every feature-gated CLI/TUI dependency

## Notes

`cargo clippy` still reports the existing reversed-empty-range lint in `src/cli/tui/ui/home_chart.rs`; this change does not touch that code.

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -76,8 +76,26 @@ jobs:
       - name: Assert CLI dependencies are absent from the library graph
         working-directory: src-tauri
         run: |
+          cli_dependency_names='clap clap_complete colored comfy-table console crossterm env_logger indicatif inquire minisign-verify ratatui self-replace semver shlex tachyonfx tar unicode-width which windows-sys winreg'
+          cli_dependency_pattern="^($(printf '%s' "$cli_dependency_names" | tr ' ' '|')) v"
+
+          cargo tree --target all -e normal --depth 1 --prefix none > default-direct-dependencies.txt
+          for cli_dependency in $cli_dependency_names; do
+            if ! grep -E "^${cli_dependency} v" default-direct-dependencies.txt > /dev/null; then
+              printf 'Known CLI dependency %s was not found in the default graph.\n' "$cli_dependency"
+              exit 1
+            fi
+          done
+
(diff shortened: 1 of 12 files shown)


bytemain:
@SaladDay 你好，非常感谢开发这么棒的一个项目，我们在基于 cc-switch-cli 封装一个 napi-rs 的包装层，方便 Node.js 调用，napi-rs 这层就可以不依赖于 cli feature（可以减小包体积），希望官方可以考虑下 cli feature optional 的这个 PR。

SaladDay:
Maintainer follow-up work is continuing in #422 because GitHub did not grant the current maintainer credential write access to this fork branch, despite maintainer edits being enabled. The original commits and authorship are preserved in the replacement branch. I will leave this PR open until the replacement CI completes.

SaladDay:
@cindyrui71 @bytemain Thank you for the contribution and for documenting the napi-rs use case. The library-only support is now available on `main`. Existing users keep the default CLI build, while embedding consumers can use `default-features = false`; Linux and Windows checks now protect that boundary.

We appreciate the thoughtful contribution!

SaladDay:
@cindyrui71 @bytemain Thank you for the contribution and for explaining the napi-rs use case. This has now been merged.

The `cli` feature remains enabled by default for existing users, while library consumers can opt out with `default-features = false`. We also added Linux and Windows checks to protect this boundary.

Thanks again for helping improve cc-switch-cli!

PR #438, opened 2026-09-04
fix(proxy): ignore empty legacy function_call placeholders

### LINKED ISSUE #437: fix(proxy): ignore empty legacy function_call in OpenAI-to-Anthropic streams
## Summary

When a Claude provider uses `apiFormat: openai_chat`, an OpenAI-compatible upstream may include an empty legacy `function_call` object in the terminal streaming chunk:

```json
{"delta":{"function_call":{"name":"","arguments":""}},"finish_reason":"stop"}
```

cc-switch CLI converts this placeholder into an Anthropic `tool_use` block with an empty ID and name. Claude Code then sends an invalid `tool_result` for that empty tool call and may continue making requests after the assistant response has finished.

## Environment

- cc-switch CLI: 5.10.4
- Provider API format: `openai_chat`
- Client route: Anthropic `/v1/messages`
- Upstream route: OpenAI-compatible `/v1/chat/completions`
- Streaming usage enabled through `stream_options.include_usage`

## Minimal upstream stream

```text
data: {"id":"chatcmpl_1","model":"compatible-model","choices":[{"delta":{"content":"ok","function_call":null}}]}

data: {"id":"chatcmpl_1","model":"compatible-model","choices":[{"delta":{"function_call":{"name":"","arguments":""}},"finish_reason":"stop"}],"usage":{"prompt_tokens":1,"completion_tokens":1}}

data: [DONE]
```

## Actual output

```text
event: content_block_start
data: {"type":"content_block_start","index":1,"content_block":{"type":"tool_use","id":"","name":"","input":{}}}

event: content_block_delta
data: {"type":"content_block_delta","index":1,"delta":{"type":"input_json_delta","partial_json":""}}

event: content_block_stop
data: {"type":"content_block_stop","index":1}
```

The message still ends with `stop_reason: "end_turn"`, so this is not a real tool call.

## Expected output

The empty legacy `function_call` placeholder should be ignored. Text content, `end_turn`, usage, and `message_stop` should be preserved without emitting a `tool_use` block.

## Root cause

In `src-tauri/src/proxy/providers/streaming.rs`, the legacy `function_call` branch checks whether `name` or `arguments` is `Some`, but does not check whether their strings are non-empty. The branch also emits a hard-coded empty tool-use ID.

## Suggested fix

Enter the legacy function-call block only when either `name` or `arguments` contains a non-empty string. Preserve the existing path for valid legacy calls and add a regression test for the terminal empty-object case.

u1544096979:
## Summary

- ignore empty legacy `function_call` placeholders during OpenAI Chat to Anthropic SSE conversion
- preserve existing handling when either the legacy function name or arguments contains data
- add a regression test for the terminal empty-object case

## Tests

- `cargo test --manifest-path src-tauri/Cargo.toml empty_legacy_function_call_does_not_emit_tool_use`
- `cargo test --manifest-path src-tauri/Cargo.toml legacy_function_call_stream_emits_tool_use_block_and_argument_delta`
- `cargo test --manifest-path src-tauri/Cargo.toml proxy::providers::streaming::tests` (21 passed)
- live `/v1/messages` stream against the affected OpenAI-compatible provider: text, `end_turn`, and `message_stop` preserved; empty `tool_use` count is zero

Closes #437

### DIFF
--- a/src-tauri/src/proxy/providers/streaming.rs
+++ b/src-tauri/src/proxy/providers/streaming.rs
@@ -481,7 +481,16 @@ pub fn create_anthropic_sse_stream(
                                     legacy_function_name = Some(name.clone());
                                 }
 
-                                if function_call.name.is_some() || function_call.arguments.is_some() {
+                                let has_legacy_function_payload = function_call
+                                    .name
+                                    .as_deref()
+                                    .is_some_and(|name| !name.is_empty())
+                                    || function_call
+                                        .arguments
+                                        .as_deref()
+                                        .is_some_and(|arguments| !arguments.is_empty());
+
+                                if has_legacy_function_payload {
                                     if let Some(index) = current_non_tool_block_index.take() {
                                         let event = json!({
                                             "type": "content_block_stop",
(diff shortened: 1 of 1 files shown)


SaladDay:
Please run `cargo fmt` from `src-tauri/` and push the changes to fix the failing formatting check.

PR #265, opened 2026-06-09
fix(codex): use relative filename for catalog path and include archived sessions

thedavidweng:
### 1. model_catalog_json uses absolute path

`set_codex_model_catalog_json_field` wrote the full absolute path (e.g. `/home/user/.codex/cc-switch-model-catalog.json`) into `config.toml`. This breaks portability and is inconsistent with how Codex itself references catalog files.

**Fix:** Write the relative filename `cc-switch-model-catalog.json` instead. The guard logic already handles both formats defensively.

### 2. Archived sessions not discoverable

The Session Manager only scanned `sessions/` but not `archived_sessions/`, making archived Codex sessions invisible in TUI/CLI.

**Fix:** Scan `archived_sessions/` alongside `sessions/` in `scan_sessions()`. Note: the session usage sync code already scanned both directories — this gap was only in the session manager.

## Changes

- `src-tauri/src/codex_config.rs` — write relative filename in `set_codex_model_catalog_json_field`
- `src-tauri/src/session_manager/providers/codex.rs` — add `archived_sessions/` scanning in `scan_sessions()`

Closes #260

PR #454, opened 2026-09-11
fix(codex): respect user-owned model_catalog_json when generating catalog

### LINKED ISSUE #260: fix(codex): model_catalog_json uses absolute path and archived sessions not discoverable
### 1. `model_catalog_json` writes absolute path instead of relative filename

In `src-tauri/src/codex_config.rs` (line ~532), `set_codex_model_catalog_json_field` writes the full absolute path (e.g. `/home/user/.codex/cc-switch-model-catalog.json`) into `config.toml` instead of the relative filename `cc-switch-model-catalog.json`.

While the guard logic in `resolve_cc_switch_catalog_path` already handles both absolute and relative paths defensively, the writing side should use the relative filename for portability and consistency with how Codex itself references catalog files.

**Location:** `src-tauri/src/codex_config.rs` — line 532: should write `CC_SWITCH_CODEX_MODEL_CATALOG_FILENAME` instead of `path.to_string_lossy().as_ref()`.

### 2. Codex archived sessions not visible in session browser

The Session Manager for Codex (`src-tauri/src/session_manager/providers/codex.rs`) only scans the `sessions/` directory. Archived sessions in `archived_sessions/` are not discoverable in the TUI or CLI session browser.

Note: the session *usage sync* code (`src/services/session_usage_codex.rs`) already scans both directories correctly — this gap is only in the session manager that powers the UI listing.

**Location:** `src-tauri/src/session_manager/providers/codex.rs` — `scan_sessions()` function (line ~24) should also walk `archived_sessions/` alongside `sessions/`.

## Expected Behavior

- `model_catalog_json` in `config.toml` should use the relative filename `cc-switch-model-catalog.json`
- Archived Codex sessions should appear in the TUI/CLI session browser alongside active sessions

czfhhh:
## Summary

Port the user-owned model-catalog fix from the desktop upstream (farion1231/cc-switch#6087). When model mapping generates the Codex catalog, cc-switch-cli currently replaces any existing `model_catalog_json` value with `cc-switch-model-catalog.json`. This PR makes the write side preserve user-managed catalog pointers.

## Problem

`set_codex_model_catalog_json_field` in `src-tauri/src/codex_config.rs` has two arms:

- The `None` arm already checks ownership and only removes cc-switch-owned pointers.
- The `Some` arm unconditionally writes `CC_SWITCH_CODEX_MODEL_CATALOG_FILENAME`, clobbering user paths such as `~/.codex/model-catalog.local.json`.

Starting the Codex local route with model mapping enabled therefore rewrites `config.toml`, breaking setups that depend on a user-managed catalog file. This builds on #265, which closed #260 by switching the same write to a relative filename.

## Fix

Mirror the `None` arm's ownership rule in the `Some` arm: claim the pointer only when it is absent or already cc-switch-owned (matched by filename); otherwise leave the user's catalog path untouched. The generated `cc-switch-model-catalog.json` file is still written, so model-mapping based proxy routing keeps working.

## Testing

- `cargo fmt --check` passes.
- `cargo test --lib`: 4334 passed, 0 failed, 2 ignored (CI-style sandbox environment).
- Added two unit tests covering the `Some` arm:
  - `set_catalog_json_some_preserves_user_owned_catalog` (absolute custom path)
  - `set_catalog_json_some_preserves_user_owned_relative_filename` (bare custom filename)
- Manually verified: with model mapping enabled and a user-owned `model_catalog_json`, enabling the Codex local route keeps the pointer in `config.toml` and multi-model routing continues to work.

### DIFF
--- a/src-tauri/src/codex_config.rs
+++ b/src-tauri/src/codex_config.rs
@@ -898,8 +898,23 @@ fn set_codex_model_catalog_json_field(
 
     match catalog_path {
         Some(_path) => {
-            // 使用相对文件名而非绝对路径，保持与 Codex 自身引用方式一致
-            doc["model_catalog_json"] = toml_edit::value(CC_SWITCH_CODEX_MODEL_CATALOG_FILENAME);
+            // Only claim the pointer when it is absent or already cc-switch-owned.
+            // A user-managed external catalog file (custom filename or path) is
+            // left untouched, mirroring the None arm's ownership rule that
+            // `resolve_cc_switch_catalog_path` relies on.
+            let is_cc_switch_owned = doc
+                .get("model_catalog_json")
+                .and_then(|item| item.as_str())
+                .map(|path| {
+                    Path::new(path).file_name().and_then(|name| name.to_str())
+                        == Some(CC_SWITCH_CODEX_MODEL_CATALOG_FILENAME)
+                })
+                .unwrap_or(true);
+            if is_cc_switch_owned {
+                // 使用相对文件名而非绝对路径，保持与 Codex 自身引用方式一致
+                doc["model_catalog_json"] =
+                    toml_edit::value(CC_SWITCH_CODEX_MODEL_CATALOG_FILENAME);
+            }
         }
         None => {
             let should_remove = doc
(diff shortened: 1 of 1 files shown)


SaladDay (approved):
LGTM

PR #462, opened 2026-09-16
fix(tui): preserve help sentinel identity in optimized builds

bryango:
**TLDR-zh_cn:** 下面这个 `std::ptr::fn_addr_eq` 在 optimized builds 中是不可靠的，导致测试失败和显示问题；可以通过 no inline 和 black_box 提升其可靠性

https://github.com/SaladDay/cc-switch-cli/blob/5ac9eac8fabf30f8d694797772588a668cc800dc/src-tauri/src/cli/tui/keymap.rs#L68-L85

## Problem

While building cc-switch-cli 5.10.5 in a Nix sandbox on Darwin, the optimized test suite fails in:

```text
cli::tui::app::tests::tests::codex_usage_rebuild_is_help_only

assertion failed:
usage::help_items(&codex, &data)
    .iter()
    .any(|(key, _)| *key == "R")
```

`R` is intentionally help-only: it should be absent from the compact key bar but present in the full help for Codex.

## Analysis

`help_items` distinguishes bindings hidden from help from help-only bindings by comparing the `shown` function pointer against the `never` sentinel:

```rust
std::ptr::fn_addr_eq(binding.shown, never as PredicateFn)
```

Both `never` and `help_only` return `false`. In optimized builds, the compiler may merge equivalent functions, causing the two sentinels to share an address. `help_items` then mistakes `help_only` for `never` and drops the binding from the full help.

We observe this in an optimized Nix sandbox build on Darwin.

## Changes

Keep the existing sentinel-based design, but make `never` distinct in optimized builds:

* mark `never` with `#[inline(never)]`
* return `std::hint::black_box(false)` from `never`
* document why the sentinel must remain distinct from `help_only`

This keeps the change local and leaves the existing binding tables and help logic unchanged.


SaladDay:
LGTM, THANKS

PR #463, opened 2026-09-16
test(skills): skip non-UTF8 filename test on macOS

bryango:
**TLDR(zh_cn):** 这个 test 不能在macos上正常运行，因为 macos 根本不支持这样的文件名

## Problem

`migration_tree_hash_preserves_non_utf8_names` creates filenames containing invalid UTF-8 bytes with `OsStringExt`.

In a Nix sandbox on Darwin, creating these test fixtures fails with `EILSEQ` before the migration hash logic is exercised. The test is useful on Unix platforms that support these filenames, but the fixture is not portable to macOS.

## Change

Run the test on Unix except macOS:

```rust
#[cfg(all(unix, not(target_os = "macos")))]
```

No production code changes.


SaladDay:
LGTM, THANKS!!

bryango:
Thank you! Here is another minimal bug fix / 另一个小小的 bug 修复:
- https://github.com/SaladDay/cc-switch-cli/pull/462

